PR #15, opened 2022-12-15
docs: add manifesto

achingbrain:
Some high-level thoughts behind the design rationale.

### DIFF
--- a/MANIFESTO.md
+++ b/MANIFESTO.md
@@ -20,28 +20,28 @@ The use cases for a distributed filesystem are incredibly far reaching and the "
 For example:
 
 - Applications deployed to browsers may wish to limit the size of the final bundle by omitting features
-- Others run in extremely adversarial environments often wish to limit the number of dependencies in the tree to limit the possibility of supply chain effects
+- Other applications may be deployed in extremely adversarial environments, and should limit the number of dependencies (throughout the dependency tree) to reduce the opportunities for supply chain attacks
 - The user should not have to include the code for features their application does not use
 
-The core of Helia will be very basic, targetting use as a library - just [js-libp2p], a [blockstore], [js-bitswap] and a posix-like API which will be extendable to add additional features such as IPNS, an RPC-API, etc.
+The core of Helia will be very focused on use as a library: just [js-libp2p], a [blockstore], [js-bitswap] and a POSIX-like API which will be extendable to add additional features such as IPNS, an RPC-API, etc.
 
 A "get you started" bundle with some common components will be provided but users are very much encouraged to roll their own version of Helia to suit their use case.
 
 ## BYO Filesystem
 
-The default filesystem for IPFS is [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md), but UnixFS has several limitations.  Support for some common unix file attributes such as mode and mtime landed in 1.5 but this has yet to make it to [kubo].
+The default filesystem for IPFS is [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md), but UnixFS has several limitations. Support for some common Unix file attributes such as mode (permission bits) and `mtime` landed in UnixFSv1.5, but this has yet to make it to [kubo].
 
-Several features are still missing such as arbitrary metadata or versioning, but they have been implemented by other filesystems such as [WNFS](https://guide.fission.codes/developers/webnative/file-system-wnfs).
+Several features are still missing from 1.5, such as arbitrary metadata (extended attributes) or versioning, but they have been implemented by other filesystems such as [WNFS](https://guide.fission.codes/developers/webnative/file-system-wnfs).
 
 That these missing features are being implemented by other filesystems is incredibly exciting and will unlock new use cases that are not possible today, so Helia will not bless any one filesystem as the One True Implementation, instead it will present an abstraction of posix filesystem operations (`ls`, `cat`, etc) as an API but the underlying filesystem(s) will be configurable.
 
 ## JavaScript first
 
-In the beginning there were node [streams](https://nodejs.org/api/stream.html#readable-streams), then there were [pull streams](https://www.npmjs.com/package/pull-stream), and finally browsers have [native streams too](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
+In the beginning there were Node.js [streams](https://nodejs.org/api/stream.html#readable-streams). Then there were [pull streams](https://www.npmjs.com/package/pull-stream). And finally, browsers have [native streams too](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
 
-Similarly [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)s can be used where [EventEmitter](https://nodejs.org/api/events.html#class-eventemitter) were previously.
+Similarly, [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)s can be used where [EventEmitter](https://nodejs.org/api/events.html#class-eventemitter)s were previously.
 
-There are many other cases where primitives that existed only in node or were provided by a popular library have now made it in to the language itself - these constructs should be used in preference to userland modules to remove dependencies and make the API types more idiomatic and recognisable as modern JavaScript.
+There are many other cases where primitives that existed only in Node.js or were provided by a popular library have now made it in to the language itself. These constructs should be preferred versus userland modules, in order to remove dependencies and make the API types more idiomatic and recognizable as modern JavaScript.
 
 This also makes it easier to support other runtimes like [deno](https://deno.land/) or [bun](https://bun.sh/) since runtime-specific code should be behind configuration such as the `"browser"` or `"exports"` fields in the `package.json`.
 
@@ -53,11 +53,11 @@ Some things are just faster in Node.js. The ability to run native code should no
 
 `js-ipfs` has no concept of permissions so all capabilities are available to every user. This has caused lots of problems with the HTTP-RPC-API in particular and lead to hastily implemented attempts to lock access down like limiting HTTP verbs or detecting user agents.
 
-Helia will integrate permissions at it's core, also allowing delegation of permissions via mechanisms such as [UCAN](https://ucan.xyz/)s.
+Helia will integrate permissions at its core, also allowing delegation of permissions via mechanisms such as [UCAN](https://ucan.xyz/)s.
 
 ## ESM and TypeScript
 
-Trying to write applications that scale has always been a challenge for JavaScript. Tools like TypeScript ease some of this pain so Helia will be full TypeScript and published as ESM-only to take advantage of modern runtimes and tooling.
+Trying to write applications that scale has always been a challenge for JavaScript. Tools like TypeScript ease some of this pain, so Helia will be written in TypeScript and published as ESM-only to take advantage of modern runtimes and tooling.
 
 ## Non-goals
 
--- a/MANIFESTO.md
+++ b/MANIFESTO.md
@@ -1,0 +1,74 @@
+# Manifesto <!-- omit in toc -->
+
+Helia aims to be a next generation IPFS implementation that takes the learnings of [js-ipfs] and applies them to a modern, modular, and efficient codebase.
+
+## Table of Contents <!-- omit in toc -->
+
+- [Modular](#modular)
+- [BYO Filesystem](#byo-filesystem)
+- [JavaScript first](#javascript-first)
+  - [(runtime-specific code where it makes sense)](#runtime-specific-code-where-it-makes-sense)
+- [Permissions](#permissions)
+- [ESM and TypeScript](#esm-and-typescript)
+- [Non-goals](#non-goals)
+  - [Networking](#networking)
+
+## Modular
+
+The use cases for a distributed filesystem are incredibly far reaching and the "bundle-everything" approach of `js-ipfs`/`kubo` does not suit every application.
+
+For example:
+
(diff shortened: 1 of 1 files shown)


MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+- [Non-goals](#non-goals)
+  - [Networking](#networking)
+
+## Modular
+
+The use cases for a distributed filesystem are incredibly far reaching and the "bundle-everything" approach of `js-ipfs`/`kubo` does not suit every application.
+
+For example:
+
+- Applications deployed to browsers may wish to limit the size of the final bundle by omitting features
```
tinytb, line 22: How do applications omit features? Is it manual, automated / tree shaking, or some combination?

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Non-goals
+
+Helia is not attempting to reinvent the wheel when it comes to layers beneath the top-level API.
+
+### Networking
+
+It will use [js-libp2p] and [js-bitswap] to ensure compatibility with existing IPFS clients.
+
+[js-ipfs]: https://github.com/ipfs/js-ipfs
```
tinytb, line 70: What will the Helia APIs and other interfaces look like? What will it take to migrate from js-ipfs?

Can migrations be done in minutes, or will they take hours?

We should also mention the fact that Helia is expected to fully replace js-ipfs eventually, and js-ipfs will be deprecated.

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+  - [Networking](#networking)
+
+## Modular
+
+The use cases for a distributed filesystem are incredibly far reaching and the "bundle-everything" approach of `js-ipfs`/`kubo` does not suit every application.
+
+For example:
+
+- Applications deployed to browsers may wish to limit the size of the final bundle by omitting features
+- Others run in extremely adversarial environments often wish to limit the number of dependencies in the tree to limit the possibility of supply chain effects
```
tinytb, line 23:
```suggestion
- Other applications may be deployed in extremely adversarial environments, and should limit the number of dependencies (throughout the dependency tree) to reduce the opportunities for supply chain attacks
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+- The user should not have to include the code for features their application does not use
```
tinytb, line 24: What will be the best way for users to do this? (exclude code)

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+The core of Helia will be very basic, targetting use as a library - just [js-libp2p], a [blockstore], [js-bitswap] and a posix-like API which will be extendable to add additional features such as IPNS, an RPC-API, etc.
```
tinytb, line 26:
```suggestion
The core of Helia will be very focused on use as a library: just [js-libp2p], a [blockstore], [js-bitswap] and a POSIX-like API which will be extendable to add additional features such as IPNS, an RPC-API, etc.
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+A "get you started" bundle with some common components will be provided but users are very much encouraged to roll their own version of Helia to suit their use case.
```
tinytb, line 28: Can we link to an example of how other libraries encourage users to roll their own versions?

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+## BYO Filesystem
+
+The default filesystem for IPFS is [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md), but UnixFS has several limitations.  Support for some common unix file attributes such as mode and mtime landed in 1.5 but this has yet to make it to [kubo].
```
tinytb, line 32:
```suggestion
The default filesystem for IPFS is [UnixFS](https://github.com/ipfs/specs/blob/main/UNIXFS.md), but UnixFS has several limitations. Support for some common Unix file attributes such as mode (permission bits) and `mtime` landed in UnixFSv1.5, but this has yet to make it to [kubo].
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+Several features are still missing such as arbitrary metadata or versioning, but they have been implemented by other filesystems such as [WNFS](https://guide.fission.codes/developers/webnative/file-system-wnfs).
```
tinytb, line 34:
```suggestion
Several features are still missing from 1.5, such as arbitrary metadata (extended attributes) or versioning, but they have been implemented by other filesystems such as [WNFS](https://guide.fission.codes/developers/webnative/file-system-wnfs).
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+That these missing features are being implemented by other filesystems is incredibly exciting and will unlock new use cases that are not possible today, so Helia will not bless any one filesystem as the One True Implementation, instead it will present an abstraction of posix filesystem operations (`ls`, `cat`, etc) as an API but the underlying filesystem(s) will be configurable.
```
tinytb, line 36:
```suggestion
That these missing features are being implemented by other filesystems is incredibly exciting and will unlock new use cases that are not possible today. Therefore, Helia will not bless any one filesystem as the One True Implementation; instead it will present an abstraction of POSIX filesystem operations (`ls`, `cat`, etc) as an API, while the underlying filesystem(s) will be configurable.
```

tinytb: Is there a list of filesystem operations that will have abstractions in the API? Are there operations that don't make sense to include? `cp`, `mv`, `rm`, etc

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+## JavaScript first
+
+In the beginning there were node [streams](https://nodejs.org/api/stream.html#readable-streams), then there were [pull streams](https://www.npmjs.com/package/pull-stream), and finally browsers have [native streams too](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
```
tinytb, line 40:
```suggestion
In the beginning there were Node.js [streams](https://nodejs.org/api/stream.html#readable-streams). Then there were [pull streams](https://www.npmjs.com/package/pull-stream). And finally, browsers have [native streams too](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+Similarly [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)s can be used where [EventEmitter](https://nodejs.org/api/events.html#class-eventemitter) were previously.
```
tinytb, line 42:
```suggestion
Similarly, [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)s can be used where [EventEmitter](https://nodejs.org/api/events.html#class-eventemitter)s were previously.
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+There are many other cases where primitives that existed only in node or were provided by a popular library have now made it in to the language itself - these constructs should be used in preference to userland modules to remove dependencies and make the API types more idiomatic and recognisable as modern JavaScript.
```
tinytb, line 44:
```suggestion
There are many other cases where primitives that existed only in Node.js or were provided by a popular library have now made it in to the language itself. These constructs should be preferred versus userland modules, in order to remove dependencies and make the API types more idiomatic and recognizable as modern JavaScript.
```

tinytb: Sorry to change the correct British spelling, but it bugged me that my browser's spellchecker marked it. Feel free to reject suggestion 😆

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+### (runtime-specific code where it makes sense)
+
+Some things are just faster in Node.js. The ability to run native code should not be underestimated so where the JS implementation of an algorithm is a proven performance bottleneck and a native version exists, it should be used.
+
+## Permissions
+
+`js-ipfs` has no concept of permissions so all capabilities are available to every user. This has caused lots of problems with the HTTP-RPC-API in particular and lead to hastily implemented attempts to lock access down like limiting HTTP verbs or detecting user agents.
+
+Helia will integrate permissions at it's core, also allowing delegation of permissions via mechanisms such as [UCAN](https://ucan.xyz/)s.
```
tinytb, line 56:
```suggestion
Helia will integrate permissions at its core, also allowing delegation of permissions via mechanisms such as [UCAN](https://ucan.xyz/)s.
```

tinytb: > integrate permissions at its core

How?

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+## ESM and TypeScript
+
+Trying to write applications that scale has always been a challenge for JavaScript. Tools like TypeScript ease some of this pain so Helia will be full TypeScript and published as ESM-only to take advantage of modern runtimes and tooling.
```
tinytb, line 60:
```suggestion
Trying to write applications that scale has always been a challenge for JavaScript. Tools like TypeScript ease some of this pain, so Helia will be fully TypeScript and published as ESM-only to take advantage of modern runtimes and tooling.
```

MANIFESTO.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Non-goals
+
+Helia is not attempting to reinvent the wheel when it comes to layers beneath the top-level API.
```
tinytb, line 64: Can we briefly explain what these layers are? Anything beyond js-libp2p and js-bitswap?

tinytb (changes requested):

MANIFESTO.md
tinytb: Required: 
- libp2p
- Somewhere to get/put blocks (blockstore)

bitswap would check blockstore first and look elsewhere if a file isn't there.

Optional feature example: IPNS (needs libp2p and pubsub)
- ~Configure IPNS service to extend Helia export?~ This would bring us back to where we are now
- Change the IPNS service to take the components you use to build a Helia node? (preferred, we think)

Could pass things to Helia that you want to use. If you don't want a feature, just don't add the module.

MANIFESTO.md
tinytb: Have features implemented as separate modules or services; if you don't use the feature, you don't use its module.

Example: libp2p has a DHT implementation (kad-dht module). If you include it in your bundle, you get the DHT implementation; if you don't include it, you don't get the code for it either

MANIFESTO.md
tinytb: Consider: Express and its middleware

MANIFESTO.md
tinytb: notes: right now, no easy way to change a DAG. If you add a file with a CID root for the folder, and you want to add a new root for the folder, it's really hard; you have to use the object API, or the DAG API directly.

Or add the CID in your MFS, then modify it that way. You need a way to make any mutation to the filesystem.

- See all operations in the MFS (non-exhaustive; other operations could be added)
- Helia will make it as easy as it is when using MFS now

MANIFESTO.md
tinytb: notes: currently, no permissions in the RPC API, so you have to be careful if your computer is exposed to the internet. Need RPC API so that you can communicate with it when it's a background process.

With UCANs, you can delegate permissions to other actors. There are reasonable use cases for the outside world to access the API.

default: require localhost to grant the initial token. On initial startup (for fresh install), Helia will first generate a token for you as the administrator. Perhaps generate it offline in a secure environment. In the future, all requests require that token, or a token with permissions delegated by that token. At that point, it's OK to be accessible to the world because an access token will be required. Thus, by default, access is limited.

Restricting to localhost doesn't stop attacks where the attacker gets the user to open a malicious script in their browser, which accesses the local node.

MANIFESTO.md
tinytb: includes: ipld, unixfs

We are designing a filesystem API. People complain the current API is too big (many methods; does a lot of things; object API, configuration API, bootstrap API)

Helia is focused on delivering an ergonomic filesystem API - not changing the other components.

Non-goal: Rearchitecting everything.

- FILES.md is the really useful thing; [files API](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-types/src/files/index.ts)
- The other things won't be part of the Helia API

tinytb: - don't need an API for config; you should just config your thing
- the other APIs are useful but don't need to be everywhere all the time. 
- e.g. IPNS can be its own thing; you would use the IPNS module and pass it libp2p and blockstore. That way, it's not part of the Helia API because it doesn't necessarily have anything to do with the filesystem; it's just mutable pointers to CIDs.
- e.g. IPLD is its own thing; use the codecs directly, but pass it the blockstore so you have a place to put the blocks.

We'll create some documentation to show how to do network IPLD operations.

MANIFESTO.md
tinytb: - depends on what parts of the API you're using. If it's just add and cat, then it's not a big change.

open question: How will users set their configuration?

note: These are ideas but they might change based on what we learn as Helia is developed.

Currently, there are many ways to set configuration. We want there to be 1 way to do it.

One big change for migrators is that they'll need to use some underlying libraries directly. Example:
- if you want to ping, you'll use libp2p directly. It won't be a method on the root of Helia.
- pubsub methods (currently just chain through to libp2p) won't be in Helia.

([js-ipfs root API](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-types/src/root.ts))

- [ ] Will need to create migration documentation: from X method before, to Y thing now.

There will be a way to do what you did with js-ipfs, but it may be a little different.

tinytb (approved):

PR #30, opened 2023-02-05
fix: feat/initial-implementation

whizzzkid:
fixes the node semver


achingbrain (changes requested):
Thanks for opening this but it's the wrong place to fix it.

This section is set by the `aegir check-project` command and the changes here will be overwritten the next time that command is run.

Instead please see [this comment](https://github.com/ipfs/helia/pull/17#discussion_r1095460604) for a link to where to open a PR to ensure the value is correct even after `aegir check-project` is run again.

achingbrain:
I'm going to close this PR as it only updates the node versions in the package.json files but is now full of conflicts because the PR it targeted has been squashed during the merge into main - it'll be simpler to open a new PR once ipfs/aegir#1184 is merged rather than try to fix the conflicts in this one.

Issue #28, opened 2023-01-27
Pinning and Garbage Collection


achingbrain:
## Context

Garbage collection in js-ipfs originally followed the go-ipfs model whereby pins were stored in a big DAG that was traversed to work out which blocks could be deleted and which couldn't while running garbage collection.

https://github.com/ipfs/js-ipfs/pull/2771 changed that to store the pins in the datastore instead of a big DAG which yielded a massive speed up when adding new pins, but garbage collection was still slow because the algorithm has to walk every dag that's pinned to build up a list of blocks in those dags.

Helia gives us an amazing opportunity to solve that slow garbage collection problem, this would be incredibly valuable to pinning services, for example, who typically don't garbage collect anything as their blockstores are so large the time it takes to run gc makes it impractical to do so.

## Gotchas

* Js-ipfs GC is a stop-the-world model, e.g. the blockstore cannot be used while GC occurs in case `add` operations add blocks that GC then immediately deletes. Coming up with a clever way to not have to do this would be greatly appreciated
* Two CIDs with different versions and/or codecs can have the same multihash - if both are pinned the removal of one pin should not delete the blocks for the other
* If the application crashes while creating a pin, it should mark the pin as `failed`
* Manually deleting blocks from the blockstore should be prevented when the block being deleted is part of a pinned DAG
* Datastore keys are used for storing pin metadata - these can be stored on filesystems so all keys should be case-insensitive

## Interface

An interface to the pinning system might look like this (somewhat similar to js-ipfs):

```ts
import { CID } from 'multiformats/cid'
import type { AbortOptions } from '@libp2p/interfaces'

enum PinStatus {
  /**
   * All blocks in the pin have been stored in the blockstore
   */
  pinned = 'pinned',

  /**
   * The pin is being created, blocks in the DAG are still being fetched from
   * the network
   */
  pending = 'pending',

  /**
   * Not all blocks could be fetched from the network - this is usually because
   * abort signal passed into the `pin.add` operation emitted it's `abort` event.
   */
  failed = 'failed'
}

interface AddOptions extends AbortOptions {
  /**
   * When pinning a DAG, Helia will ensure that all blocks in the DAG are present in
   * the blockstore which may involve network operations. By default Helia will traverse
   * the entire DAG but pass a depth here to limit that behaviour.
   */
  depth?: number

  /**
   * A user-chosen name for the pin
   */
  name?: string

  /**
   * User-specific metadata for the pin
   */
[...]

PR #36, opened 2023-02-20
feat: add pinning API

achingbrain:
Adds the pinning API as specified in #28 - see that issue for discussion

Benchmarks incoming!

Closes: #28

### DIFF
--- a/packages/helia/package.json
+++ b/packages/helia/package.json
@@ -142,14 +142,21 @@
     "@libp2p/interface-libp2p": "^1.1.0",
     "@libp2p/interfaces": "^3.3.1",
     "blockstore-core": "^3.0.0",
+    "cborg": "^1.10.0",
     "interface-blockstore": "^4.0.1",
     "interface-datastore": "^7.0.3",
     "interface-store": "^3.0.4",
     "ipfs-bitswap": "^16.0.0",
+    "it-all": "^2.0.0",
+    "it-drain": "^2.0.0",
     "it-filter": "^2.0.0",
     "it-merge": "^2.0.0",
     "it-pushable": "^3.1.2",
-    "multiformats": "^11.0.1"
+    "mortice": "^3.0.1",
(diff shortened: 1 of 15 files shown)


achingbrain:
I've updated the benchmark to omit setup and teardown from the timed sections, though it needs tinylibs/tinybench#33 to be merged before it can be run.

The benchmark creates a load of small pinned dags and a load of unpinned blocks, then runs garbage collection.

In the results below we start with 1,000 dags/unpinned blocks, up to 10,000, increasing in increments of 1,000, so we start with 4k blocks (1k DAGs of 3 blocks each + 1k unpinned blocks) and end with 40k blocks.

It tests helia, js-ipfs and kubo though I've omitted js-ipfs from the graphs because it skewed them so far to make the helia/kubo series be plotted as flat lines at the bottom of the graph.  As an indication running gc with 40k blocks took js-ipfs 5.5 hours, it takes helia about 4 seconds(!).  The difference here is that js-ipfs walks the DAG of each block in the blockstore to build up a list of unpinned blocks whereas helia already knows which blocks are pinned due to the reference counting.

I ran these benchmarks on my laptop while doing other things which is why there is some wobble in the lines so I've added trend lines to make things a bit clearer.

image

---

Splitting the setup out of the timed function also allowed me to benchmark the different steps separately for some bonus stats.

This is deleting pins (start with deleting 1k pins, end with 10k pins):

image

Adding blocks (kubo adds blocks with the `ipfs.block.put` RPC command, helia uses the blockstore `putMany` function, which allows the blockstore to control concurrency):

image

Finally adding new pins to the pinset, (start with adding 1k pins, end with 10k pins, the ms value is how long adding that many pins takes).  Helia scales better because it's not mutating a giant DAG that gets bigger with every pin added:

image

SgtPooki:
@achingbrain , this is great. nice work

achingbrain:
Vive le JS!

Issue #35, opened 2023-02-20
2023/2024 Contributor Hospitality Ideas


BigLep:
Below are items that stuck out when coming to https://github.com/ipfs/helia with fresh eyes imagining what it's like for someone who hears about Helia, wants to learn more, and wants to contribute potentially:

## README Items
- [x] Reference helia-unixfs, helia-ipns, etc.  We don't give pointers to these important modules. https://github.com/ipfs/helia/pull/76
- [x] How do I get started as a consumer?  Reference https://github.com/ipfs-examples/helia-examples
- [x] Reference demo-slides, recordings, etc: https://github.com/ipfs/helia/pull/80
- [x] Remove erroneous verbiage: "The core of IPFS is the [Files API](https://github.com/ipfs/js-ipfs/blob/master/docs/core-api/FILES.md), which will likewise be implemented in Helia. These initial building blocks are in development now; have a look at this repo's PR(s)."  Is this true.  Isn't this in helia-unixfs?
- [x] @whizzzkid : Fix API docs links: https://github.com/ipfs/aegir/pull/1228
- [x] Link to code coverage.  [The badge](https://codecov.io/gh/ipfs/helia) lists it as unknown.  This is a new endeavor.  We should be able to show off good numbers to start. https://github.com/ipfs/aegir/pull/1195
- [x] @BigLep Have a section about who some of our users are.  I expect we don't have any currently.  That's ok, but lets solicit for people to share us.  We want a "your logo here": https://github.com/ipfs/helia/pull/83
- [x] @BigLep Update project status section: https://github.com/ipfs/helia/pull/83
- [x] Answer "Where should a user ask questions?"
- [x] @achingbrain Answer "What is the relationship of this project to js-ipfs?"
- [x] @achingbrain Answer "How do we prove interoperability with other IPFS implementations like Kubo?"
- [x] @achingbrain Answer "How does it perform compared to other implementations including js-ipfs?"
- [ ] Answer "What runtimes do we support (node versions, browser distributions/versions including mobile/desktop)"
- [ ] I think we could use more visuals on the readme.  I think having something visual showing the relationship between ipfs/helia, js-libp2p, and the other Helia repos.  I think seeing this helps drive home the point that Helia is modular.

## Manifesto Items
[...]

PR #78, opened 2023-04-05
chore: Adds codeowners

whizzzkid:
Addresses: #35 

Adding @ipfs/helia-dev as CODEOWNERS.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @ipfs/helia-dev


achingbrain (changes requested):
Click on "Files changed".  It says "This CODEOWNERS file contains errors".

They need to be addressed before this can be merged.

achingbrain (approved):

BigLep:
@whizzzkid : a few things here:
1. We need this for the other Helia projects (ipns, unixfs) and idelia helia-examples (although I know it's a different org)
2. You can actually declare CODEOWNERS in github-mgmt.  That will allow you to have one PR that has the codeowners in all the helia examples (rather than doing a PR per repo).  Here's an example PR where that was done: https://github.com/multiformats/github-mgmt/pull/67 (your call of course)

PR #79, opened 2023-04-05
fix: :arrow_up: Min node version 18.

whizzzkid:
Relates: #35 

Removes:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'semantic-release@20.1.3',
npm WARN EBADENGINE   required: { node: '>=18' },
npm WARN EBADENGINE   current: { node: 'v16.18.1', npm: '8.19.2' }
npm WARN EBADENGINE }
```

Since package-lock files are not shipped, incompatible packages can show up on older versions of node. There is also some weirdness around switching node versions between builds, so keeping it latest LTS can help.


achingbrain:
This change needs to be made in `aegir` otherwise it will be overwritten on the next run of `npx aegir check-project`.

Please PR these lines: https://github.com/ipfs/aegir/blob/15f01ea4f27738a9483a464552286e0739d797b1/src/check-project/utils.js#L252-L253

PR #95, opened 2023-04-16
fix(types): Add missing types

whizzzkid:
In this PR adding missing return types.

### DIFF
--- a/packages/helia/src/storage.ts
+++ b/packages/helia/src/storage.ts
@@ -175,7 +175,7 @@ export class BlockStorage implements Blocks {
       const storage = this
 
       options.onProgress?.(new CustomProgressEvent('blocks:delete-many:blockstore:delete-many'))
-      yield * this.child.deleteMany((async function * (): AsyncIterable<CID> {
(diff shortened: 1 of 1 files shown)


packages/helia/src/storage.ts
```diff
@@ -174,7 +175,7 @@ export class BlockStorage implements Blocks {
       const storage = this
 
       options.onProgress?.(new CustomProgressEvent('blocks:delete-many:blockstore:delete-many'))
-      yield * this.child.deleteMany((async function * () {
+      yield * this.child.deleteMany((async function * (): AsyncIterable<CID> {
```
achingbrain, line 178:
```suggestion
      yield * this.child.deleteMany((async function * (): AsyncGenerator<CID> {
```

achingbrain (approved):
LGTM, just a small type change required.

Interesting that [this rule](https://github.com/ipfs/eslint-config-ipfs/blob/main/src/ts.js#L20) didn't pick this up before.  Any ideas why?

PR #97, opened 2023-04-18
docs: add usage example

autonome:


README.md
```diff
@@ -11,6 +11,18 @@
...
+
+```
+import { createHelia } from 'helia';
+
+const ipfs = await createHelia();
+
+const myImmutableAddress = ipfs.add(JSON.stringify({ foo: "bar" }));
+
+console.log(ipfs.get(myImmutableAddress));
+```
```
achingbrain, line 24: I think this has a bunch of assumptions baked in.

- What is the CID version?
- What is the CID codec?
- How does the string get turned into a block?
- How does the block get turned back into a string?
- What happens if the string is bigger than the block size limit?

These are not questions a new user should have to answer, so we could encapsulate them into another module?

```suggestion
import { createHelia } from 'helia'
import { strings } from '@helia/strings'

const helia = await createHelia()
const str = strings(helia)

const myImmutableAddress = await str.add(JSON.stringify({ foo: "bar" }))

console.log(await str.get(myImmutableAddress));
```

The next step would be to ask the user what if you don't want strings?  How about JSON in, JSON out?  Here's `@ipld/dag-json` or maybe wrap it in `@helia/json`?

What do you think?

PR #127, opened 2023-05-19
feat: provide default libp2p instance

### LINKED ISSUE #121: Provide a default libp2p instance
## Done Criteria
Helia can be constructed without a libp2p instance.  In this case, a "good default" libp2p instance is provided.

## Why Important
libp2p exposes a lot of knobs.  Knowing the right one sets for a good connectivity story is not cear to new users.  This reduces the barrier to entry and also serves as a base set to potentially tweak.

## Notes
This default libp2p instance should be well commented so that interested users can learn from it to determining how they might follow suit or tweak.

achingbrain:
The original intention was to allow users to configure a libp2p node to their requirements but it's a non-trivial undertaking unless the user is deeply familiar with the libp2p stack.

Instead, let's provide node and browser libp2p instances with sensible defaults that give the user the best chance of success on first try.

Fixes #121

### DIFF
--- a/packages/helia/package.json
+++ b/packages/helia/package.json
@@ -145,6 +145,7 @@
     "@ipld/dag-pb": "^4.0.3",
     "@libp2p/bootstrap": "^8.0.0",
     "@libp2p/interface-libp2p": "^3.1.0",
+    "@libp2p/interface-pubsub": "^4.0.1",
     "@libp2p/interfaces": "^3.3.2",
     "@libp2p/ipni-content-routing": "^1.0.0",
     "@libp2p/kad-dht": "^9.3.3",
--- a/packages/helia/package.json
+++ b/packages/helia/package.json
@@ -138,18 +138,29 @@
     "release": "aegir release"
   },
   "dependencies": {
+    "@chainsafe/libp2p-gossipsub": "^8.0.0",
+    "@chainsafe/libp2p-noise": "^12.0.0",
+    "@chainsafe/libp2p-yamux": "^4.0.2",
     "@helia/interface": "^1.0.0",
-    "@ipld/dag-pb": "^4.0.2",
-    "@libp2p/interface-libp2p": "^1.1.0",
-    "@libp2p/interfaces": "^3.3.1",
-    "@libp2p/logger": "^2.0.5",
+    "@ipld/dag-pb": "^4.0.3",
+    "@libp2p/bootstrap": "^8.0.0",
+    "@libp2p/interface-libp2p": "^3.1.0",
+    "@libp2p/interfaces": "^3.3.2",
+    "@libp2p/ipni-content-routing": "^1.0.0",
+    "@libp2p/kad-dht": "^9.3.3",
+    "@libp2p/logger": "^2.0.7",
+    "@libp2p/mdns": "^8.0.0",
+    "@libp2p/mplex": "^8.0.3",
+    "@libp2p/tcp": "^7.0.1",
+    "@libp2p/webrtc": "^2.0.4",
+    "@libp2p/webtransport": "^2.0.1",
     "blockstore-core": "^4.0.0",
     "cborg": "^1.10.0",
     "datastore-core": "^9.0.0",
     "interface-blockstore": "^5.0.0",
     "interface-datastore": "^8.0.0",
     "interface-store": "^5.0.1",
-    "ipfs-bitswap": "^17.0.0",
+    "ipfs-bitswap": "^18.0.0",
     "it-all": "^3.0.1",
     "it-drain": "^3.0.1",
     "it-filter": "^3.0.1",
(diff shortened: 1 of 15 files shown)


packages/helia/package.json
```diff
@@ -162,18 +174,19 @@
     "uint8arrays": "^4.0.3"
   },
   "devDependencies": {
-    "@chainsafe/libp2p-noise": "^11.0.0",
-    "@chainsafe/libp2p-yamux": "^3.0.5",
     "@ipld/dag-cbor": "^9.0.0",
     "@ipld/dag-json": "^10.0.1",
-    "@libp2p/websockets": "^5.0.3",
+    "@libp2p/websockets": "^6.0.1",
```
SgtPooki, line 179: should this be "optional" dependencies?

SgtPooki (approved):
lgtm, just a question.

packages/helia/package.json
SgtPooki: same for others that are browser only?

achingbrain: No, optional dependencies means "if this dependency failed to install, carry on anyway".

I think what you are saying is "let the user decide whether to install this dep or not", which is what "peerDependencies" used to mean, now npm installs peer deps whether you like it or not so I'm not sure there's a way to do that.

packages/helia/src/utils/libp2p.browser.ts
```diff
@@ -0,0 +1,68 @@
...
+      circuitRelayTransport({
+        discoverRelays: 1
+      })
+    ],
+    connectionEncryption: [
+      noise()
+    ],
+    streamMuxers: [
+      yamux(),
+      mplex()
```
BigLep, line 41: can we kill off mplex?

packages/helia/src/utils/libp2p.ts
```diff
@@ -0,0 +1,72 @@
...
+      circuitRelayTransport({
+        discoverRelays: 1
+      })
+    ],
+    connectionEncryption: [
+      noise()
+    ],
+    streamMuxers: [
+      yamux(),
+      mplex()
```
BigLep, line 45: can we remove?

packages/helia/src/utils/libp2p.ts
```diff
@@ -0,0 +1,72 @@
...
+    ],
+    peerDiscovery: [
+      mdns(),
+      bootstrap({
+        list: [
+          '/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN',
+          '/dnsaddr/bootstrap.libp2p.io/p2p/QmQCU2EcMqAqQPR2i9bChDtGNJchTbq5TbXJJ16u19uLTa',
+          '/dnsaddr/bootstrap.libp2p.io/p2p/QmbLHAnMoJPWSCR5Zhtx6BHJX9KiKNN6tpvbUcqanj75Nb',
+          '/dnsaddr/bootstrap.libp2p.io/p2p/QmcZf59bWwK5XFi76CZX8cbJ4BhTzzA3gU1ZjYZcYW3dwt'
+        ]
```
BigLep, line 55: This is duplicated.  Maybe factor it out, and also add comments on where this list comes from?

packages/interface/src/index.ts
```diff
@@ -28,11 +27,11 @@ export type { Await, AwaitIterable } from 'interface-store'
 /**
  * The API presented by a Helia node.
  */
-export interface Helia {
+export interface Helia<T = Libp2p> {
```
BigLep, line 30: You don't have to spend time on this, but just for my own learning, I'm curious why we had to parameterize Helia.

BigLep (commented):
Are there are any docs or examples we need to update?

packages/helia/src/utils/libp2p.browser.ts
achingbrain: Maybe - do we have stats for how well used yamux is yet?

js-ipfs never shipped with it included so you'd lose compatibility with any existing deployments.

packages/interface/src/index.ts
achingbrain: Libp2p is parameterised by the service map, so the generic is added here to let us reflect that in the instance of Helia returned from the factory.

packages/helia/src/utils/libp2p.browser.ts
BigLep: actually, I was wanting to make sure we don't proliferate mplex, but as long as we prefer yamux that is the key thing (which you're doing).  I'm not aware of harm of one also offering mplex, and assuming there are no downsides to it, I'm game to include it for compatibility.  

That said, if go-libp2p ships without mplex by default, I think we'd be fine to remove it.  I don't have stats here and I don't think we need to block to find that out.

packages/helia/src/utils/libp2p.ts
achingbrain: https://github.com/ipfs/helia/pull/129

packages/helia/src/utils/libp2p.browser.ts
achingbrain: Muxers are negotiated in-order so yamux will be preferred over mplex by any peer that supports both.

PR #128, opened 2023-05-19
fix: add helia version to agent version

### LINKED ISSUE #122: Helia identifies itself to the network
## Done Criteria
Helia identifies itself to the network by default.  Specifically this mean:
* libp2p identity - will be identifiable from DHT bootstrappers for example.  Kubo does this.
* user-agent header when making HTTP calls - This is relevant for example when making calls to cid.contact

We want to also get the Helia version as well (e.g., “helia/vX.Y.Z”) just as Kubo does.   

We should only set the "user agent" if a user hasn't specified an override one. 

## Why Important
This enables Helia's prevalence to be detected/measured.  This is important for getting signal on Helia's adoption in production, which in turn affects things like:
1. funding and resourcing decisions
2. client considerations for network/protocol upgrades

## Notes
1. At the end of this, problab should be able to generate the Helia equivalent of https://probelab.io/ipfsdht/#active-kubo-versions

achingbrain:
If the user has configured identify but not overridden the agent version, add helia and its version number to the string.

Fixes #122

### DIFF
--- a/packages/helia/src/index.ts
+++ b/packages/helia/src/index.ts
@@ -21,10 +21,12 @@
  * ```
  */
 
+import { logger } from '@libp2p/logger'
 import { MemoryBlockstore } from 'blockstore-core'
 import { MemoryDatastore } from 'datastore-core'
 import { HeliaImpl } from './helia.js'
 import { createLibp2p } from './utils/libp2p.js'
+import { name, version } from './version.js'
 import type { Helia } from '@helia/interface'
 import type { Libp2p } from '@libp2p/interface-libp2p'
 import type { PubSub } from '@libp2p/interface-pubsub'
@@ -33,6 +35,8 @@ import type { Blockstore } from 'interface-blockstore'
 import type { Datastore } from 'interface-datastore'
 import type { CID } from 'multiformats/cid'
 import type { MultihashHasher } from 'multiformats/hashes/interface'
+
+const log = logger('helia')
 
 /**
  * DAGWalkers take a block and yield CIDs encoded in that block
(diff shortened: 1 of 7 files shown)


packages/helia/src/version.ts
```diff
@@ -0,0 +1,2 @@
+export const version = '0.0.0'
+export const name = 'helia'
```
SgtPooki, line 2: why not have this file be a json file

SgtPooki (approved):
lgtm but confused why we're saving version & name in `version.ts` instead of `version.json` or a more aptly named `agentInfo.json`

packages/helia/src/version.ts
achingbrain: Node 18 still shows a warning if you import json from js:

```
(node:25810) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
```

packages/helia/test/factory.spec.ts
```diff
@@ -35,4 +38,50 @@ describe('helia factory', () => {
     await helia.datastore.put(key, block)
     expect(await helia.datastore.has(key)).to.be.true()
   })
+
+  it('adds helia details to the AgentVersion', async () => {
+    helia = await createHelia()
+
+    const peer = await helia.libp2p.peerStore.get(helia.libp2p.peerId)
+    const agentVersionBuf = peer.metadata.get('AgentVersion')
+    const agentVersion = new TextDecoder().decode(agentVersionBuf)
```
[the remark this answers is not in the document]
achingbrain: The AgentVersion is used by the identify protocol and it's not something people generally interact with directly.

packages/helia/src/index.ts
```diff
@@ -122,5 +126,49 @@ export async function createHelia (init: HeliaInit = {}): Promise<Helia<unknown>
...
+
+  if (versionBuf == null) {
+    // identify was not configured
+    return
+  }
+
+  let versionStr = new TextDecoder().decode(versionBuf)
+
+  if (versionStr.match(/js-libp2p\/\d+\.\d+\.\d+\sUserAgent=/) == null) {
+    // the user changed the agent version
```
BigLep, line 157: I'm not saying we need to open a new PR, but I think it would be nice in future for these kinds of things if we linked to the source of the information (presumably somewhere in js-libp2p).

BigLep (commented):
I know there are tests/aseerts here (great - thanks).  For my own learning, could you please paste what the resulting full "versionStr" is that Helia will now be using (and what it was before).  (I know I could get things up and running to figure out myself - I'm asking as a time saver for me - thank you!)

packages/helia/src/index.ts
achingbrain: Agree, but it’s buried fairly deeply in the identify service so would need a change there - it’s not something that’s needed to be accessed before.

achingbrain:
Before:

```
libp2p/x.x.x UserAgent=$USER_AGENT
```

Where `$USER_AGENT` is either the current node.js version (e.g. `v18.16.0`), or the browser's user agent string (e.g. `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36`)

After:

```
helia/x.x.x libp2p/x.x.x UserAgent=$USER_AGENT
```

BigLep:
Thanks for the info @achingbrain .  I added ann FAQ entry: https://github.com/ipfs/helia/wiki/%E2%9D%93-FAQ#-how-does-helia-identify-itself. Please modify if I got anything wrong.

PR #135, opened 2023-05-24
fix: dedupe bootstrap list

achingbrain:
List of bootstrap nodes is in one place now and links to the source.


SgtPooki (approved):
might be worth updating https://github.com/ipfs/helia/wiki/Migrating-from-js-IPFS#name-ipns to indicate that config isn't manually necessary now (but still useful to show what's needed)

PR #167, opened 2023-06-28
test: create add-dir benchmark

SgtPooki:
Creates a new benchmark that compares adding directories with helia, js-ipfs, and kubo (via ipfsd-ctl) 

### Related issues
* https://discuss.ipfs.tech/t/helia-add-directory/16618


SgtPooki:
> @SgtPooki I created: #169
> 
> However both the baseline and fix benchmark numbers don't align on my machine, can we also create a benchmark runner github action to generate consistent benchmarks?

what do you mean they don't align? They are different on your machine than mine? That's to be expected.

As far as a github action for running these benchmarks and keeping them posted somewhere.. I think that's an *excellent* idea and something we should create an issue for (see https://github.com/ipfs/helia/issues/170). It would be nice to output those benchmarks in line with the version of helia used when creating them so we can see how things change as we progress.

SgtPooki:
ping @achingbrain

achingbrain:
@SgtPooki I made some changes to this PR in https://github.com/ipfs/helia/pull/171 - let me know what you think

benchmarks/add-dir/src/index.ts
```diff
@@ -0,0 +1,182 @@
...
+      }
+    }
+    console.table(suite.tasks.map(({ name, result }) => {
+      if (result?.error != null) {
+        return {
+          Implementation: name,
+          'ops/s': 'error',
+          'ms/op': 'error',
+          runs: 'error',
+          stddev: 'error',
```
SgtPooki, line 162:
```suggestion
          p99: 'error',
```

SgtPooki:
@achingbrain are you good with merging this?

PR #231, opened 2023-08-14
feat!: iterable pinning

saul-jb:
This PR reworks the pinning to be async generators instead of promises with support for batching, granting users more fine-grained control of the pinning/downloading of blocks.

As expected using a higher batch number gives much better performance than the old pinning system since its doing more fetching in parallel. A batch number of 1 will replicate the old system of a simple queue.

This is a basic example of throttling the pins download, more advanced systems can be made to control how many large pins can be downloaded at once with various speeds by calling `next` on the iterables.
```javascript
// Pin the first 10 blocks of the CID.
for await (const cids of helia.pins.add(cid, { batch: 10 }) {
  // Wait for a second before pinning the next 10 blocks.
  await new Promise(resolve => setTimeout(resolve, 1000)
}

// The CID is now pinned.
```

Another example, altering the batch size:
```javascript
const itr = helia.pins.add(cid, { batch: 8 })

// Pin the first 8 blocks.
await itr.next()

// Pin the next 5 blocks.
await itr.next(5)
```

Adding pins:
```javascript
// Before:
await helia.pins.add(cid)

// After:
await all(helia.pins.add(cid))
```

Removing pins:
```javascript
// Before:
await helia.pins.rm(cid)

// After:
await all(helia.pins.rm(cid))
```

Getting the pinned value:
```javascript
// Before:
const pin = await helia.pins.add(cid)

// After:
const itr = helia.pins.add(cid)

let next = await itr.next()

while (!next.done) {
  next = await itr.next()
}

const pin = next.value
```


saul-jb:
It might be desirable to change the default behavior of the first iteration call to allow it to pull all the local blocks without a batch limit. This would allow pins to quickly catch up to where they were if the program was interrupted without having to await the next iteration call, giving the user an easy way to have 'resume download' type behavior.

Something like this:
```javascript
const itr = helia.pins.add(cid, { syncLocal: true })[Symbol.asyncIterator]()

// Fetches all the blocks that are in the local blockstore not limited to batch numbers:
const localCids = await itr.next()

// Fetches the next blocks from the network according to the batch number:
const downloadedCids = await itr.next()
```

achingbrain:
This is interesting, but I wonder if we couldn't simplify the internal implementation by just having it be a generator that pins blocks as fast as possible, then if the user wants to batch them up (or parallelise, or parallel batch) they can use existing modules like [it-batch](https://www.npmjs.com/package/it-batch), [it-parallel](https://www.npmjs.com/package/it-parallel), [it-parallel-batch](https://www.npmjs.com/package/it-parallel-batch) or a combination of things from [streaming-iterables](https://www.npmjs.com/package/streaming-iterables)?

That would give them a lot more control over the performance characteristics of pinning, and we'll have a simpler codebase to maintain.

saul-jb:
> That would give them a lot more control over the performance characteristics of pinning, and we'll have a simpler codebase to maintain.

I agree and have updated the code.

Adding pins:
```javascript
// Before:
await helia.pins.add(cid)

// After:
await all(parallel(helia.pins.add(cid), { concurrency: 1 }))
```

Removing pins:
```javascript
// Before:
await helia.pins.rm(cid)

// After:
await all(parallel(helia.pins.rm(cid), { concurrency: 1 }))
```

Getting the pinned value:
```javascript
// Before:
const pin = await helia.pins.add(cid)

// After:
const itr = helia.pins.add(cid)

let output = await itr.next()

while (output.done === false) {
  await output.value()
  output = await itr.next()
}

const pin = next.value
```

One thing this is still missing that I would like to see is being able to continue a pin where it left off or catch up without rate limiting pinning of local blocks. The simple use case that illustrates this would be the rate limited pinning of a large DAG and the program crashes part way and you want it to continue where it left off.

Do you think a `skipLocal` option is the best way to implement this or is there a better way?

packages/helia/src/pins.ts
```diff
@@ -38,15 +35,20 @@ interface DatastorePinnedBlock {
...
+/**
+ * Callback for updating a {@link DatastorePinnedBlock}'s properties when
+ * calling `#updatePinnedBlock`
+ *
+ * The callback should return `false` to prevent any pinning modifications to
+ * the block, and true in all other cases.
+ */
+interface WithPinnedBlockCallback {
+  (pinnedBlock: DatastorePinnedBlock): boolean
+}
```
SgtPooki, line 47: @achingbrain I pulled this out into a locally defined interface and added some context

packages/helia/src/pins.ts
```diff
@@ -84,38 +86,24 @@ export class PinsImpl implements Pins {
...
+    for await (const promise of this.#walkDag(cid, depth, options)) {
+      yield async () => {
+        const cid = await promise()
 
-    // if a job in the queue errors, throw that error
-    const deferred = defer()
+        await this.#updatePinnedBlock(cid, (pinnedBlock: DatastorePinnedBlock) => {
+          // do not update pinned block if this block is already pinned by this CID
+          if (pinnedBlock.pinnedBy.find(c => uint8ArrayEquals(c, cid.bytes)) != null) {
+            return false
```
SgtPooki, line 96: Note: we return false to say that we do not want to process the pinning of this content any further.

packages/helia/src/pins.ts
```diff
@@ -84,38 +86,24 @@ export class PinsImpl implements Pins {
...
+          }
 
-    queue.on('error', (err): void => {
-      queue.clear()
-      deferred.reject(err)
-    })
+          pinnedBlock.pinCount++
+          pinnedBlock.pinnedBy.push(cid.bytes)
+          return true
```
SgtPooki, line 101: we should be returning true by default inside a `WithPinnedBlockCallback`

packages/helia/src/pins.ts
```diff
@@ -179,7 +185,10 @@ export class PinsImpl implements Pins {
       }
     }
 
-    withPinnedBlock(pinnedBlock)
+    const shouldContinue = withPinnedBlock(pinnedBlock)
+    if (!shouldContinue) {
+      return
+    }
```
SgtPooki, line 191: @achingbrain @saul-jb this is how I've implemented the `skipLocal` option that was mentioned. We already had the logic to skip content if it was pinned, we just needed a way to exit early.

All tests pass, but I'm not sure of any implications of returning prior to checking for `pinnedBlock.pinCount === 0`. However, since only `.add` is returning false when `pinnedBlock.pinnedBy.find(c => uint8ArrayEquals(c, cid.bytes)) != null`, and everything else returns true, I believe this should continue working as it had previously.

packages/helia/test/pins.recursive.spec.ts
```diff
@@ -85,4 +87,35 @@ describe('pins (recursive)', () => {
       }
     }
   })
+
+  it('can resume an interrupted pinning operation', async () => {
```
SgtPooki, line 91: Added a test to simulate an interrupted pinning event by only iterating over it partially, and then attempting to pin from the root again.

SgtPooki (approved):
Self review, plus approving prior code. Should be looked at again by @saul-jb and @achingbrain

packages/helia/test/pins.recursive.spec.ts
SgtPooki: We could optimize this further, but a worthy optimization would likely require a drastic change to the pinning response and input types. 

We could try to prevent calls to `#updatePinnedBlock` entirely by asking users to provide cids to skip (which would have been returned in the first `.add` call). But `pinnedBy` is checking the block content explicitly with `uint8ArrayEquals`, not just CID strings, so I assume this would cause more work in cases where we're not resuming.

Open to ideas

SgtPooki:
dismissed my review and pointed to regression seen when updating benchmarks/gc in https://github.com/ipfs/helia/pull/346

packages/helia/src/pins.ts
```diff
@@ -131,39 +119,57 @@ export class PinsImpl implements Pins {
...
-      return
-    }
+  async * #walkDag (cid: CID, maxDepth: number, options: AbortOptions): AsyncGenerator<() => Promise<CID>> {
+    const queue: Array<() => Promise<CID>> = []
+    const promises: Array<Promise<CID>> = []
 
-    const dagWalker = this.dagWalkers[cid.code]
+    const enqueue = (cid: CID, depth: number): void => {
+      queue.push(async () => {
+        const promise = Promise.resolve().then(async () => {
```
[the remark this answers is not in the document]
saul-jb: That is just another way to write `;(async () => {})()` without the need for a semi colon.

packages/helia/src/pins.ts
```diff
@@ -131,39 +119,57 @@ export class PinsImpl implements Pins {
...
+      const func = queue.shift()
+
+      if (func == null) {
+        await promises.shift()
+
+        continue
+      }
+
+      yield func
+    }
```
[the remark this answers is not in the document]
saul-jb: Yes, when the queue is exhausted but there are still promises running we wait until the next one has resolved before trying to pull more from the queue.

achingbrain:
I'm going to merge this so we have the API changes in and the [v3](https://github.com/ipfs/helia/issues/337) release is unblocked, we can address the performance characteristics in a follow-up.

The existing "resumable pinning" change is useful though it's more "restartable pinning" than "resumable" - I'll open another issue to discuss.

achingbrain (approved):

PR #232, opened 2023-08-14
feat: re-export types from @helia/interface

achingbrain:
`@helia/interface` lets us depend on the shape of a `Helia` without having to depend on `helia` directly which is useful when you are writing something lightweight and don't want to pull in libp2p etc during an `npm i`, but in some cases we will have a dep on `helia` so re-export the types from `@helia/interface` to not force people to have a dependency on both.

### DIFF
--- a/packages/helia/src/index.ts
+++ b/packages/helia/src/index.ts
@@ -37,6 +37,12 @@ import type { Libp2pOptions } from 'libp2p'
 import type { CID } from 'multiformats/cid'
 import type { MultihashHasher } from 'multiformats/hashes/interface'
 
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):

achingbrain:
> I thought this was intentionally left out to segregate the responsibilities

It was, but from a practical perspective it's a pain to use.

By shipping `@helia/interface` as well as `helia`, the user has the option to, for example, depend on `@helia/interface` in their production code and then supply an implementation from `helia` in their test code, but if their production code depends on `helia` anyway we might as well save them the extra dep.

PR #251, opened 2023-09-02
fix(kubo): ⬆️ Upgrading go-ipfs to kubo

whizzzkid:
In this PR:

- replaced go-ipfs with kubo
- remove relevant `//@ts-expect-error no types`


SgtPooki (approved):
Lgtm

PR #252, opened 2023-09-02
docs: CNAME record for helia.js.org

whizzzkid:
In this PR:
- adding CNAME record for `helia.js.org`
- Doc: https://typedoc.org/options/output/#cname
- js.org: https://js.org/?helia
- Corresponding js.org PR: https://github.com/js-org/js.org/pull/8473


achingbrain:
I thought we'd decided to go with a different domain name.  Do we need this?

PR #267, opened 2023-09-18
fix: try circuit relay transport first

achingbrain:
Make circuit relay the first transport to try so it will match addresses ahead of supporting transports.

E.g. where the address is `/<webtransport>/<circuit-relay>` get the circuit relay transport to dial first as it will strip off the relay part, hand control back to the transport manager and then run the circuit relay protocol over the newly created webtransport connection.

This is a tactical fix for a bug that needs to be addressed in libp2p.

### DIFF
--- a/packages/helia/src/utils/libp2p-defaults.ts
+++ b/packages/helia/src/utils/libp2p-defaults.ts
@@ -7,6 +7,7 @@ import { type DualKadDHT, kadDHT } from '@libp2p/kad-dht'
 import { mdns } from '@libp2p/mdns'
 import { mplex } from '@libp2p/mplex'
 import { tcp } from '@libp2p/tcp'
+import { webRTC, webRTCDirect } from '@libp2p/webrtc'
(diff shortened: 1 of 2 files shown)


SgtPooki (approved):

packages/helia/src/utils/libp2p-defaults.ts
```diff
@@ -35,15 +36,18 @@ export function libp2pDefaults (): Libp2pOptions<DefaultLibp2pServices> {
...
     transports: [
-      tcp(),
-      webSockets(),
       circuitRelayTransport({
         discoverRelays: 1
-      })
+      }),
+      tcp(),
+      webRTC(),
+      webRTCDirect(),
```
SgtPooki, line 49: Any idea when we would add webtransport here?

achingbrain: When it's supported in node - see https://github.com/libp2p/js-libp2p-webtransport/pull/31

Issue #274, opened 2023-10-02
Update Helia to fallback to trustless HTTP gateways


SgtPooki:
discussed in https://pl-strflt.notion.site/Helia-reliable-retrieval-technical-design-golden-path-255-15c4d3c25a404a85b6db8bf3f8d1f310?pvs=4

Should do https://github.com/ipfs/helia/issues/272 first

```[tasklist]
### Tasks
- [x] create ByteProvider interface
- [x] add `byteProviders` config key to Helia configuration object
- [x] Emit onProgress events that specify which provider (dht, gateway url, etc..) is returning bytes
- [x] implement default fallback byteProvider for a trustless http gateway
- [x] clarify with @achingbrain whether we need to verify bytes or if this is already handled
- [ ] https://github.com/ipfs/helia/pull/276
- [ ] https://github.com/ipfs/helia/pull/280
- [ ] https://github.com/ipfs/helia/pull/281
- [ ] https://github.com/ipfs/helia/pull/284
- [ ] https://github.com/ipfs/helia/pull/285
- [ ] https://github.com/ipfs/helia/issues/287
```

SgtPooki:
https://github.com/ipfs/helia/pull/276 is still in progress, but I could use some input from @achingbrain that the direction i'm heading in works before I flesh everything out

SgtPooki:
some of this work was moved to https://github.com/ipfs/helia/pull/280 and also https://github.com/ipfs/helia/pull/281 on top of that

SgtPooki:
The work here is done and merged into the `main` branch but is not released on `@latest` tag on npm yet. You can experiment with the changes made on the `@next` tag.

BigLep:
@SgtPooki : what are the conditions for closing this?
1. Helia release?
2. Any additional docs we need to add? (or is that already in the API docs)?  (If already in the API docs, can you point me to the relevant ones so I can look at them from a user's point of view?)
3. ?

Thanks

SgtPooki:
> @SgtPooki : what are the conditions for closing this?
> 
> 1. Helia release?

Yep. I discussed this with Alex and we talked about how we would let this sit in `@next` tagged release and request devs try it out for a bit before merging it. There's no guarantee that devs will try it, but I *plan* to.

> 2. Any additional docs we need to add? (or is that already in the API docs)?  (If already in the API docs, can you point me to the relevant ones so I can look at them from a user's point of view?)

[...]

PR #276, opened 2023-10-04
feat: helia supports additional byteProviders

SgtPooki:
Work for addressing issues in https://github.com/ipfs/helia/issues/274.

The direction I'm thinking with this is:

* We have ByteProvider and BlockProviders
* Users are given a helper function to create trustless-http-gateway byte-providers 
* Users provide byteProviders array in helia instantiation (which will be used to construct BlockProviders that are passed a blockstore, and cannot be edited)
* NetworkedStorage, which is what calls bitswap, was slightly refactored to call both bitswap & any given blockProviders, wrapping and listening to the given signal so that we can cancel any outstanding requests and prevent additional requests if any byteProvider returns quickly enough
* Users have no control over BlockProvider instances (because a blockstore is required), but shouldn't need to (correct me if i'm wrong)

---

This is still a WIP, but confirmed it's working locally with:

```javascript
import { createHelia } from 'helia'

const node = await createHelia()

const blocks = await node.blockstore.get(CID.parse('QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D'), {signal: AbortSignal.timeout(10000)})
console.log('done')
```

Output:

```bash
  helia:gateway-block-provider getting block for QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D from https://dweb.link +0ms
  helia:gateway-block-provider getting block for QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D from https://cf-ipfs.com +2ms
  helia:gateway-block-provider getting block for QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D from https://4everland.io +0ms
  helia:gateway-block-provider getting block for QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D from https://w3s.link +1ms
  helia:gateway-block-provider getting block for QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D from https://cloudflare-ipfs.com +1ms
  helia:gateway-block-provider got block for QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D from https://dweb.link +229ms
```

---


packages/helia/src/utils/block-provider.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Get a block by cid, using the given ByteProvider
+   */
+  async get (cid: CID, options: GetOfflineOptions & AbortOptions): Promise<Uint8Array> {
+    if (options.offline !== true && !(await this.blockstore.has(cid))) {
+      try {
+        const block = await this.#provider.get(cid, options)
+
+        await this.blockstore.put(cid, block, options)
```
SgtPooki, line 43: we could wrap the blockstore on put that re-hashes it to verify.

we don't need to parse the data.. we just need to verify that when we hash the block with the cid.multihash.code is the same. 

add method to this class for verifying bytes. might be able to use util methods in blockstore classes/package

achingbrain:
This is great!  I've taken some of the ideas here and opened https://github.com/ipfs/helia/pull/280 which I think is a bit simpler.  Let me know what you think.

PR #280, opened 2023-10-09
feat: configurable block providers

achingbrain:
Adds configurable block providers to allow using bitswap but also other methods such as trustless gateways and any yet-to-be-invented way of resolving a CID to a block.

There are two block providers out of the box - bitswap and a trustless gateway provider that tries to download the block from one of a set of preconfigured gateways.

The block is verified after the provider has provided it.  This lets users call `helia.blockstore.put` without verification for performance but anything coming from the network will be verified.


packages/helia/src/block-providers/bitswap-block-provider.ts
```diff
@@ -0,0 +1,59 @@
...
+            return hasher.code === codecOrName || hasher.name === codecOrName
+          })
+
+          if (hasher != null) {
+            return hasher
+          }
+
+          throw new Error(`Could not load hasher for code/name "${codecOrName}"`)
+        }
+      }
```
achingbrain, line 33: Bitswap already verifies blocks pulled from the network.  If we like the approach in this PR we should PR bitswap to skip verification and let it be handled here.

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.gateways = urls.map(url => new URL(url.toString()))
+  }
+
+  notify (cid: CID, block: Uint8Array, options?: ProgressOptions): void {
+    /* no-op */
+  }
+
+  async get (cid: CID, options: AbortOptions & ProgressOptions<TrustlessGatewayGetBlockProgressEvents> = {}): Promise<Uint8Array> {
+    // choose a gateway
+    const url = this.gateways[Math.floor(Math.random() * this.gateways.length)]
```
achingbrain, line 32: Randomly chosen gateway.  If a gateway returns a 5xx we may wish to temporarily remove it from rotation and try another?

packages/interface/src/blocks.ts
```diff
@@ -61,3 +62,18 @@ ProgressOptions<DeleteBlockProgressEvents>, ProgressOptions<DeleteManyBlocksProg
...
 }
+
+export interface BlockProvider<
+  NotifyProgressOptions extends ProgressOptions = ProgressOptions,
+  WantProgressOptions extends ProgressOptions = ProgressOptions
+> {
+  /**
+   * Notify a block provider that a new block is available
+   */
+  notify(cid: CID, block: Uint8Array, options?: NotifyProgressOptions): void
```
achingbrain, line 73: Right now this is only used by bitswap to fulfil want list requests from other peers.  I guess you could argue that this BlockProvider is now Providing Blocks to the network?  Or we could factor it out.

SgtPooki: I couldn't mentally reconcile a `BlockProvider` with a `notify` method. This is why I didn't pull out bitswap in my PR, but naming is hard and I think supporting generic bitswap functionality is more important than getting the naming perfect.

> I guess you could argue that this BlockProvider is now Providing Blocks to the network

This line of thought makes sense to me. We should make sure it's clear to consumers that not all blockProviders need to support a "notify" method.

packages/helia/src/index.ts
```diff
@@ -91,6 +96,12 @@ export interface HeliaInit<T extends Libp2p = Libp2p> {
    */
   dagWalkers?: DAGWalker[]
 
+  /**
+   * A list of strategies used to fetch blocks when they are not present in
```
SgtPooki, line 100: We should update the comment unless we change the name to something like `BlockStrategies` or `BlockExchangeStrategies`.

```suggestion
   * A list of providers used to fetch blocks when they are not present in
```

packages/helia/src/utils/networked-storage.ts
```diff
@@ -46,10 +68,11 @@ export class NetworkedStorage implements Blocks {
...
     }
 
-    if (this.bitswap?.isStarted() === true) {
-      options.onProgress?.(new CustomProgressEvent<CID>('blocks:put:bitswap:notify', cid))
-      this.bitswap.notify(cid, block, options)
-    }
+    options.onProgress?.(new CustomProgressEvent<CID>('blocks:put:providers:notify', cid))
+
+    this.blockProviders.forEach(provider => {
+      provider.notify(cid, block, options)
```
SgtPooki, line 74:
```suggestion
      provider.notify?.(cid, block, options)
```

or better: keep a list of blockProviders that have `notify` methods and only call those.

packages/helia/src/utils/networked-storage.ts
```diff
@@ -71,8 +94,10 @@ export class NetworkedStorage implements Blocks {
     })
 
     const notifyEach = forEach(missingBlocks, ({ cid, block }): void => {
-      options.onProgress?.(new CustomProgressEvent<CID>('blocks:put-many:bitswap:notify', cid))
-      this.bitswap?.notify(cid, block, options)
+      options.onProgress?.(new CustomProgressEvent<CID>('blocks:put-many:providers:notify', cid))
+      this.blockProviders.forEach(provider => {
+        provider.notify(cid, block, options)
```
SgtPooki, line 99:
```suggestion
        provider.notify?.(cid, block, options)
```

or better: keep a list of blockProviders that have `notify` methods and only call those.

packages/helia/src/utils/networked-storage.ts
```diff
@@ -21,16 +25,34 @@ export interface GetOptions extends AbortOptions {
...
 
   /**
    * Create a new BlockStorage
    */
-  constructor (blockstore: Blockstore, options: BlockStorageInit = {}) {
+  constructor (blockstore: Blockstore, init: NetworkedStorageStorageInit) {
     this.child = blockstore
-    this.bitswap = options.bitswap
+    this.blockProviders = init.blockProviders ?? []
+    this.hashers = init.hashers ?? []
```
SgtPooki, line 40:
```suggestion
    this.blockProviders = init.blockProviders ?? []
    this.blockProvidersWithNotify = this.blockProviders.filter((p) => p.notify != null)
    this.hashers = init.hashers ?? []
```

packages/helia/src/utils/networked-storage.ts
```diff
@@ -21,16 +25,34 @@ export interface GetOptions extends AbortOptions {
...
+  private started: boolean
 
   /**
    * Create a new BlockStorage
    */
-  constructor (blockstore: Blockstore, options: BlockStorageInit = {}) {
+  constructor (blockstore: Blockstore, init: NetworkedStorageStorageInit) {
     this.child = blockstore
-    this.bitswap = options.bitswap
+    this.blockProviders = init.blockProviders ?? []
```
SgtPooki, line 39: Having an internal-only representation of `blockProviders`, or whatever we name it, would allow us to rank/prioritize them dynamically more easily. (i.e. `BlockProviders` wrapping `byteProviders` in my PR)

Or we can expose a priority/weight on the class where users can specify their preferences.
Or we can expose a filter/sort config method on Helia root config where users can specify strategies for de-prioritizing certain `blockProviders`

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+> {
+  private readonly gateways: URL[]
+
+  constructor (urls: string[]) {
+    this.gateways = urls.map(url => new URL(url.toString()))
+  }
+
+  notify (cid: CID, block: Uint8Array, options?: ProgressOptions): void {
+    // no-op
+  }
```
SgtPooki, line 28:
```suggestion
```

I think it would be better to have an interface that supports defining `BlockProvider` classes without a `notify` method.

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export class TrustedGatewayBlockProvider implements BlockProvider<
+ProgressOptions,
+ProgressOptions<TrustlessGatewayGetBlockProgressEvents>
+> {
+  private readonly gateways: URL[]
+
+  constructor (urls: string[]) {
+    this.gateways = urls.map(url => new URL(url.toString()))
+  }
```
SgtPooki, line 24: prevent users from having to url.toString() just for us to convert it to a url again later. The code works either way and gives consumers flexibility

```suggestion
  constructor (urls: (string | URL)[]) {
    this.gateways = urls.map(url => new URL(url.toString()))
  }
```

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
SgtPooki: we definitely should enable some ability to retry other given gateways. IIRC, we discussed in https://pl-strflt.notion.site/Helia-reliable-retrieval-technical-design-golden-path-255-15c4d3c25a404a85b6db8bf3f8d1f310?pvs=4 that just spamming all gateways would be fine for now because we're trying to optimize for successful retrieval.

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * A BlockProvider that accepts a list of trustless gateways that are queried
+ * for blocks. Individual gateways are randomly chosen.
+ */
+export class TrustedGatewayBlockProvider implements BlockProvider<
+ProgressOptions,
+ProgressOptions<TrustlessGatewayGetBlockProgressEvents>
+> {
+  private readonly gateways: URL[]
```
SgtPooki, line 20:
```suggestion
class InternalGateways {
	url: URL
	attempts: number
	successes: number
	constructor(url: string|URL) {
	  this.url = url instanceOf URL ? url : url.toString()
	}
	toString() {
	  return this.url.toString()
	}
	get successRate() {
	  if (this.attempts === 0) return 0
	  return this.successes / this.attempts
	}
}

export type TrustlessGatewayGetBlockProgressEvents =
  ProgressEvent<'trustless-gateway:get-block:fetch', URL>

/**
 * A BlockProvider that accepts a list of trustless gateways that are queried
 * for blocks. Individual gateways are randomly chosen.
 */
export class TrustedGatewayBlockProvider implements BlockProvider<
ProgressOptions,
ProgressOptions<TrustlessGatewayGetBlockProgressEvents>
> {
  private readonly gateways: InternalGateways[]
```

We should have some sort of ranking/tracking of gateway retrieval attempts so we can deprioritize gateways that are consistently failing

SgtPooki: i'll do a PR with a slight refactor in this direction

packages/helia/src/index.ts
```diff
@@ -114,6 +125,23 @@ export interface HeliaInit<T extends Libp2p = Libp2p> {
   holdGcLock?: boolean
```
SgtPooki, line 125: is this still being used?

SgtPooki (commented):
This is definitely more in the direction I wanted to go, but I wasn't sure how to reconcile bitswap as a "blockProvider". comments on that on appropriate lines.

I think a better name would help future devs and consumers, and I'm more inclined to unblock Helia-retrieval now rather than prevent Helia API difficulties in the future, but we should decide on a good name now since we’re exposing that to consumers.

A few other things I’ve thought of w.r.t. naming and future-proofing:

* We could have a more generic `BlockExchanger` interface that doesn’t imply a direction; A `BlockProvider` that implies 1. provide to others (notify) and 2. provides to you (get) is confusing to me.
* We could use naming similar to go-bitswap and js-ipfs-bitswap (“data trading module”), i.e. `BlockTrader`, `BlockBroker`
* `BlockBroker` (my preferred name) could consist of both a `BlockRetriever` (get agent) and a `BlockAnnouncer` (notify agent) interface, and NetworkedStorage should deal only in `BlockBrokers`. (this could be done in future PRs)
* We could use `BlockProviderStrategy`, `RemoteBlockStrategies`, or `BlockStrategies` to align with the description you wrote in `packages/helia/src/index.ts`

----

I'll put up a PR with some of the changes I've mentioned in comments so they can be pulled in without much edit if desired.

packages/helia/src/block-providers/bitswap-block-provider.ts
achingbrain: We'd release a major. Despite it being published for consumption outside IPFS projects realistically I don't anyone else is using it bar a few weekend hackers.

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
achingbrain: True, but I had a moment of doubt when I realised we'd be making 50 HTTP requests for a 10 block DAG.

packages/helia/src/index.ts
achingbrain: I intentionally didn't use "providers" here because of the terminology clash with DHT provider records.

packages/helia/src/utils/networked-storage.ts
achingbrain: This method is now optional on the interface.

packages/helia/src/utils/networked-storage.ts
achingbrain: This sounds like an improvement we can make if it becomes necessary in future?

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
achingbrain: Agreed, this method is now optional on the interface.

packages/helia/src/index.ts
achingbrain: Yes.

packages/helia/src/utils/networked-storage.ts
achingbrain: I think we should make this optimisation if it becomes clear this is a bottleneck?

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
achingbrain: I think the linter prefers `Array<string | URL>` to this form.

achingbrain: I've updated the signature.

packages/helia/src/block-providers/trustless-gateway-block-provider.ts
achingbrain: Sounds good.

packages/helia/src/utils/networked-storage.ts
achingbrain: Done!

SgtPooki (approved):
LGTM, thanks Alex 👍

PR #284, opened 2023-10-10
feat: BlockBroker factory support

SgtPooki:
Adds on to changes from https://github.com/ipfs/helia/pull/280 and https://github.com/ipfs/helia/pull/281

This makes it much easier to support constructing helia with custom blockBrokers while still allowing consumers easy access to the default BitswapBlockBroker. One particular use-case where this comes in handy: our tests.

Take the code in `packages/helia/test/fixtures/create-helia.ts` for example. Without these changes, you have to do something like:

```typescript
import { webSockets } from '@libp2p/websockets'
import * as Filters from '@libp2p/websockets/filters'
import { MemoryBlockstore } from 'blockstore-core' // new import
import { MemoryDatastore } from 'datastore-core' // new import
import { circuitRelayTransport } from 'libp2p/circuit-relay'
import { identifyService } from 'libp2p/identify'
import { BitswapBlockBroker } from '../../src/block-brokers/index.js' // new import
import { createHelia as createNode } from '../../src/index.js'
import { defaultHashers } from '../../src/utils/default-hashers.js' // new import
import { createLibp2p } from '../../src/utils/libp2p.js' // new import
import type { Helia } from '@helia/interface'

export async function createHelia (): Promise<Helia> {
  const datastore = new MemoryDatastore()
  const blockstore = new MemoryBlockstore()
  const libp2p = await createLibp2p(datastore, {
    addresses: {
      listen: [
        `${process.env.RELAY_SERVER}/p2p-circuit`
      ]
    },
    transports: [
      webSockets({
        filter: Filters.all
      }),
      circuitRelayTransport()
    ],
    connectionGater: {
      denyDialMultiaddr: async () => false
    },
    services: {
      identify: identifyService()
    }
  })
  return createNode({
    blockstore,
    blockBrokers: [
      new BitswapBlockBroker(libp2p, blockstore, defaultHashers())
    ],
    datastore,
    libp2p
  })
}
```

However, with these changes, we can now do this:

```typescript
import { webSockets } from '@libp2p/websockets'
import * as Filters from '@libp2p/websockets/filters'
import { circuitRelayTransport } from 'libp2p/circuit-relay'
import { identifyService } from 'libp2p/identify'
import { BitswapBlockBrokerFactory } from '../../src/block-brokers/bitswap-block-broker.js' // new import
import { createHelia as createNode } from '../../src/index.js'
import type { Helia } from '@helia/interface'

export async function createHelia (): Promise<Helia> {
  return createNode({
    blockBrokers: [
[...]

PR #281, opened 2023-10-09
feat: TrustlessGatewayBlockBroker prioritizes & tries all gateways

SgtPooki:
Slight changes on top of #280

- keep track of individual gateway status and expose reliability rating we can use to sort/prioritize
- Try all gateways in TrustlessGatewayBlockBroker, sequentially, sorted by reliability

### todo

- [x] fix tests (hasher)
- [x] Add tests for racing block providers
- [x] Add tests for prioritizion of trustlessGateways that error/succeed or are untried

PR #285, opened 2023-10-10
feat: blockBrokers can control block validation

SgtPooki:
- feat: blockBrokers can control block validation

Built on top of https://github.com/ipfs/helia/pull/281 & https://github.com/ipfs/helia/pull/284

This PR adds the `validateFn` option to `BlockRetriever.retrieve` options and slightly modifies
how `raceBlockRetrievers` inside of `NetworkedStorage` is handled. Some callouts:

* BlockBroker's are given a TRUSTED `validateFn` to call when they retrieve blocks, and MAY call this method prior to returning the block
* If they call `validateFn`, when
    * it doesn't throw, we consider the blocks to have been verified (since we own the inner `validateFn` and given MultihashHashers)
    * it throws, they can handle recourse within their `retrieve` method (e.g. try other gateways, other peers/transports/etc, whatever they want)
        * If they do not, we consider that blockBroker to have failed the `retrieve` method and other `BlockBrokers` will continue to `raceBlockRetrievers` as expected
* If they do not call `validateFn`, we will call it for them, and then
    * If it doesn't throw, we consider the blocks to have been verified (since we own the inner `validateFn` and given MultihashHashers)
    * If it throws, we consider that blockBroker to have failed the `retrieve` method and other `BlockBrokers` will continue to `raceBlockRetrievers` as expected
* The types of the `retrieve` method in `BitswapBlockBroker` were modified so a `validateFn` can be handled there, but no work towards that effort should be done until https://github.com/ipfs/js-ipfs-bitswap/issues/603 is handled.

### DIFF
--- a/packages/helia/src/utils/networked-storage.ts
+++ b/packages/helia/src/utils/networked-storage.ts
@@ -6,7 +6,7 @@ import filter from 'it-filter'
 import forEach from 'it-foreach'
 import { CustomProgressEvent, type ProgressOptions } from 'progress-events'
 import { equals as uint8ArrayEquals } from 'uint8arrays/equals'
-import type { BlockBroker, Blocks, Pair, DeleteManyBlocksProgressEvents, DeleteBlockProgressEvents, GetBlockProgressEvents, GetManyBlocksProgressEvents, PutManyBlocksProgressEvents, PutBlockProgressEvents, GetAllBlocksProgressEvents, GetOfflineOptions, BlockRetriever, BlockAnnouncer } from '@helia/interface/blocks'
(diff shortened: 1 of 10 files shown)


packages/helia/src/utils/networked-storage.ts
```diff
@@ -196,39 +196,57 @@ export class NetworkedStorage implements Blocks, Startable {
...
-          // verify block
-          const hash = await hasher.digest(block)
-
-          if (!uint8ArrayEquals(hash.digest, cid.multihash.digest)) {
-            throw new CodeError('Hash of downloaded block did not match multihash from passed CID', 'ERR_HASH_MISMATCH')
+          if (!blocksWereValidated) {
+            // the blockBroker either did not throw an error when attempting to validate the block
+            // or did not call the validateFn at all. We should validate the block ourselves
+            await validateFn(block)
           }
```
SgtPooki, line 245: We *could* have a callback method where we tell `blockBroker`s that they gave us an invalid block, and then wait for them to handle it however they see fit, but that would require us to do another try block, and ends up in some weird "back-and-forth" control scenario when blockProviders have complex retry flows. 

We don't want to get in the business of handling those retry flows, so exposing a function they can call to determine whether blocks are valid or not prior to giving control back to us is the most simple & flexible method I can think of right now.

We can adjust if we see problems with this, but I think this is fairly flexible, and won't end up with duplicate validations (except where js-ipfs-bitswap is doing so internally)

SgtPooki (commented):
self review.. things look good.

PR #290, opened 2023-10-24
docs(version): ✨ Pulling version number from helia.

whizzzkid:
Fixes the package version on the docs hosted on helia.io.

- Pulls the latest version from `helia`.
- Updates the root `package.json`
- Builds the docs.

helia-version-doc

This should fix the version in the next release.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,8 +33,9 @@
     "release": "run-s build docs:no-publish npm:release docs",
     "npm:release": "aegir exec --bail false npm -- publish",
     "release:rc": "aegir release-rc",
-    "docs": "NODE_OPTIONS=--max_old_space_size=8192 aegir docs -- --exclude packages/interop --excludeExternals",
-    "docs:no-publish": "NODE_OPTIONS=--max_old_space_size=8192 aegir docs --publish false -- --exclude packages/interop"
+    "update:version": "node scripts/update-version.js",
+    "docs": "npm run update:version && NODE_OPTIONS=--max_old_space_size=8192 aegir docs -- --exclude packages/interop --excludeExternals",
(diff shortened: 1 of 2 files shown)


achingbrain (changes requested):
Please can we be careful about the commit messages used as they will affect final version numbers of shipping software.

This is not a feature so `feat:` is inappropriate.  Instead, please choose from the [available tags](https://github.com/ipfs/helia/blob/main/.github/workflows/main.yml#L174-L179) - `docs:` seems like the obvious one to me

If necessary please review the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) documentation.

achingbrain:
This is a strange fix because v2.0.3 is the current version of [helia](https://www.npmjs.com/package/helia) which is compliant with v2.0.0 of the [helia interface](https://www.npmjs.com/package/@helia/interface).

In theory we'll see lots of patch releases of helia but very few releases to the interface.

The API docs are monorepo docs rather than specific to a given module and the version number comes from the [root package.json](https://github.com/ipfs/helia/blob/main/package.json#L3).

Perhaps there's a config key for the [root typedoc.json](https://github.com/ipfs/helia/blob/main/typedoc.json) to remove display of the monorepo version as it doesn't make much sense to the consumer?  Then we don't need contortions like the one introduced in this PR.

achingbrain:
> `typedoc.json` has `"includeVersion": false` but doesn't seem to work 🤔.

It was being hard-coded in aegir's `docs` command, this should be fixed now.

achingbrain:
Alternative: https://github.com/ipfs/helia/pull/296

achingbrain:
Closing in favour of https://github.com/ipfs/helia/pull/296

Issue #170, opened 2023-06-29
feat: run benchmarks via github action and publish results


SgtPooki:
We should have results publicly accessible that we can link users to and also see the progress of helia improvements.

### References
* https://github.com/ipfs/helia/pull/167#issuecomment-1613474257
* https://github.com/ipfs/helia/pull/167#issuecomment-1613542482
```[tasklist]
### Tasks
- [ ] test: automatically run benchmark tests after each release and publish somewhere
- [ ] test: resolve issues with out-of-order benchmark timings (https://github.com/ipfs/helia/pull/171#issuecomment-1614755422)
```

laurentsenta:
Would you like help with this? ipdx have bandwidth to help here,

An experiment running in CI and using [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark): https://github.com/laurentsenta/helia/actions/runs/5715166671/attempts/3#summary-15484283839

A rough proposal, we would:

1. Run the benchmarks in CI, as-is, so you get reproducibility in PRs (and use our self-hosted runners if needed)
2. Implement automated tooling to run matrix'es of benchmarks and triggers errors in PRs
3. Export the data to Grafana, so you get more dashboard'ing options

SgtPooki:
@laurentsenta that would be amazing! the report at https://github.com/laurentsenta/helia/actions/runs/5715166671/attempts/3#summary-15484283839 is great but we definitely should make sure it's running on nodeLTS (there's a flag about it running on node12?)

> Run the benchmarks in CI, as-is, so you get reproducibility in PRs (and use our self-hosted runners if needed)

Yes. that sounds good to me, but ultimately @achingbrain is the lead of Helia and we'd want him to review any plans we have

> Implement automated tooling to run matrix'es of benchmarks and shows stats + triggers errors in PRs

Yes definitely. Running all(2 for now) benchmarks with an output similar to what you shared with previous & current times would be amazing.

> Export the data to Grafana, so you get more dashboard'ing options

Exporting the data so we have graphs and whatnot would be great, but I want to make sure we're not increasing any maintenance burdens on Alex or others helping with IP-JS!

laurentsenta:
Great to hear; thanks for the details @SgtPooki!
Regarding node12, it's caused by an outdated caching action, the benchmark is running with lts as you expect 👍

ocdbytes:
Hey @laurentsenta 

Do you need any help with this issue?

PR #292, opened 2023-10-26
test: benchmark CI from laurentsenta

SgtPooki:
see https://github.com/laurentsenta/helia/blob/e2a79985b78429f1abf008373794197ccc2f0f76/.github/workflows/benchmark.yml

fixes https://github.com/ipfs/helia/issues/170


laurentsenta:
@SgtPooki thanks for looking into this! You might have a problem with reproducibility as is, when I experimented with this 6 runs gaves 6 different results (see the [Attempts](https://github.com/laurentsenta/helia/actions/runs/5715166671/attempts/3) button at the top), you might need self hosted runners. cc @galargh

PR #296, opened 2023-10-30
docs: update generated docs to include version in module names

achingbrain:
Updates to latest aegir version with simplified readmes, auto-copying @packageDocumentation to "About" section and includes version numbers in the module versions.

image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,4 +1,4 @@
-name: ci
+name: CI
 on:
   push:
     branches:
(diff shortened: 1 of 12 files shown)


SgtPooki (approved):
So with this change, we wont have TOC in READMEs anymore, but we get auto generated About sections based on the @packageDocumentation for packages in each subpackage of a monorepo?

achingbrain:
Yes - in the generated docs we have a TOC on the right already so it's redundant:

image

> we get auto generated About sections based on the @packageDocumentation for packages in each subpackage of a monorepo

Also yes, with the slight caveat that we have to run `npm run docs` to update it.

PR #297, opened 2023-10-30
fix: replace IPNI gateway with delegated routing client

achingbrain:
Removes proprietary IPNI content routing implementation with `@helia/delegated-routing-v1-http-api-client` which also gives us better peer routing and IPNS support.


packages/helia/src/utils/libp2p-defaults.browser.ts
```diff
@@ -54,14 +55,12 @@ export function libp2pDefaults (): Libp2pOptions<DefaultLibp2pServices> {
...
     ],
-    contentRouters: [
-      ipniContentRouting('https://cid.contact')
-    ],
     services: {
       identify: identifyService(),
       autoNAT: autoNATService(),
       pubsub: gossipsub(),
       dcutr: dcutrService(),
+      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://trustless-gateway.link'),
```
SgtPooki, line 63: @aschmahmann is this the correct link?

SgtPooki (commented):
Good with this once we confirm the endpoint

packages/helia/src/utils/libp2p-defaults.browser.ts
aschmahmann: No, that's the link for the trustless gateway (i.e. the one that conforms to https://specs.ipfs.tech/http-gateways/trustless-gateway/ and only sends back blocks and CARs, which means it's much less likely to get overzealously blocked by someone). The one for delegated routing is: delegated-ipfs.dev (see https://github.com/protocol/bifrost-community/issues/1)

achingbrain: I've updated it - it works! https://delegated-ipfs.dev/routing/v1/providers/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi

SgtPooki (approved):

PR #303, opened 2023-11-06
see ipfs.ts

1cbyc:
used "try...catch" block to handle potential errors and log them to the console for debugging. 

adjusted the import functions too.

hope the changes look good?

### DIFF
--- a/benchmarks/gc/src/ipfs.ts
+++ b/benchmarks/gc/src/ipfs.ts
@@ -1,68 +1,99 @@
-import os from 'node:os'
-import path from 'node:path'
-import { create } from 'ipfs-core'
-import all from 'it-all'
-import drain from 'it-drain'
-import type { GcBenchmark } from './index.js'
+const os = require('os');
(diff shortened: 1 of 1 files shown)


SgtPooki:
@1cbycthanks for the contribution but we should not be using require statements anymore

1cbyc:
> @1cbycthanks for the contribution but we should not be using require statements anymore

ooh, okay. that makes sense.

SgtPooki (changes requested):
I dont see the benefits of these changes, but do see a regression in maintainable code.

It would be easier to get a significantly smaller change merged that doesnt change code style or patterns.

achingbrain (changes requested):
@1cbyc thank you for taking the time to open this PR.

Could you go into a bit of detail about why these changes are necessary and what problem they solve?

achingbrain:
I'm going to close this PR, it's not clear what value the changes add and it won't run due to use of CJS in ESM.

Issue #289, opened 2023-10-23
feat: create `@helia/http` library without bitswap/libp2p deps 


SgtPooki:
Helia consumers should be able to use Helia without depending on the p2p network. Ideally, users could use Helia modularly, with trustless-http-gateway (or other source of getting blocks) and validate those blocks, using only the IPFS portion of Helia.

## Current

It doesn't seem like we can "remove" libp2p entirely, but we can disable it:

```js
const helia = await createHelia({
  blockBrokers: [
    trustlessGateway()
  ],
  libp2p: {
    start: false,
    connectionManager: {
      minConnections: 0,
    },
  }
})
```

## Future (with `@helia/http`) 

No runtime js-libp2p deps are pulled in with this module

```
import type { Helia } from '@helia/interface'
import { createHttpOnlyHelia } from '@helia/http' // export name to be decided. should probably be `createHelia`
import { unixfs } from '@helia/unixfs'

const helia: Helia<null> = createHttpOnlyHelia() 

const fs = unixfs(helia)

// ...
```

```[tasklist]
### Tasks
- [ ] https://github.com/ipfs/helia/issues/341
- [ ] create @helia/http package
- [ ] @helia/http should accept an interface of only blockstore+blockBrokers that should work like Helia's networked blockstore.
- [ ] Add `@helia/http` usage example to readme
```

achingbrain:
Maybe we should accept `false` for `libp2p` for this?

```js
const helia = await createHelia({
  blockBrokers: [
    trustlessGateway()
  ],
  libp2p: false
})
```

achingbrain:
Thinking about this a bit more, if there's no libp2p/bitswap then you're pulling in a whole lot of extra deps for zero benefit and I wonder if this is even Helia any more.

It might be another implementation of the [Helia interface](https://helia.io/modules/_helia_interface.html)?  This keeps it compatible with `@helia/unixfs` etc.

SgtPooki:
I think Helia can be Helia even without libp2p (people don't want to do p2p stuff, but they do want to do content-addressing stuff), but I agree it feels odd. 

> It might be another implementation of the [Helia interface](https://helia.io/modules/_helia_interface.html)? This keeps it compatible with @helia/unixfs etc.

@achingbrain are you thinking of going with something like 

```ts
import {type HeliaRaw} from '@helia/interface`
import {createHelia} from 'helia'

const helia = await createHelia({
  blockBrokers: [
    trustlessGateway()
  ],
[...]

Issue #344, opened 2023-12-11
Disabling libp2p doesn't work


2color:
## Background

in #289, there's an example showing how to start helia with Libp2p disabled:

```js
const helia = await createHelia({
  blockBrokers: [
    trustlessGateway()
  ],
  libp2p: {
    start: false
  }
})
```

The type for libp2p is defined here: https://github.com/libp2p/js-libp2p/blob/d994311cc9b59cdcbd5968bb7799b7fbe14a8961/packages/libp2p/src/index.ts#L125

However, when starting Helia like that, Helia still attempts to open quite a lot of libp2p connections to the bootstrappers and other nodes.

What should be the recommended configuration for this?

## Reproduction

See the example here: https://stackblitz.com/edit/github-cadjaa?file=src%2Fprovider%2FHeliaProvider.jsx

SgtPooki:
I will provide a minimal config here that shows how to do this. This is done in helia-http-gateway currently, but its got a lot of other code around it.

SgtPooki:
The dials libp2p is doing by default can be disabled by using:

```
const helia = await createHelia({
  blockBrokers: [
    trustlessGateway()
  ],
  libp2p: {
    start: false,
    connectionManager: {
      minConnections: 0,
    },
  }
})
```

but it still "does stuff". You can see the console for libp2p debug logs that I enabled with `localStorage.setItem('debug', 'libp2p*')` in `App.jsx` 

You can see that libp2p is starting, even though we set start to false. I think this is currently a bug with Helia/libp2p at the line https://github.com/ipfs/helia/blob/299bb0942bbfae492db938c4ccad4e835bab2dbd/packages/helia/src/helia.ts#L69. We should handle if users want to start libp2p, and if they pass `libp2p: {start:false}` we should prevent starting libp2p in Helia start, and require the user to manage it.

You can also do something like this:

```
        const helia = await createHelia({
          blockBrokers: [
            trustlessGateway({
              gateways: ['https://dweb.link', 'https://cf-ipfs.com'],
            }),
          ],
          libp2p: {
            start: false,
            connectionManager: {
              minConnections: 0,
            },
            services: {},
            peerDiscovery: [],
          },
        });
        await helia.libp2p.stop()
```

to silence libp2p further. I think any further work for this should be handled by https://github.com/ipfs/helia/issues/289

SgtPooki:
[...]

PR #372, opened 2024-01-09
feat!: add @helia/http to monorepo

achingbrain:
Adds code from https://github.com/meandavejustice/helia-http to the monorepo, preserving commit history.

Fixes: https://github.com/ipfs/helia/issues/289 and https://github.com/ipfs/helia/issues/344

BREAKING CHANGE: the `libp2p` property has been removed from the `Helia` interface in `@helia/interface` - it is still present on the return type of `createHelia` from the `helia` module


2color:
> I think you still want libp2p here, only configured without transports or listen addresses - instead configured with only http delegated routing clients as in https://github.com/libp2p/js-libp2p-example-delegated-routing/blob/main/src/libp2p.js - this will allow for it's use with @helia/ipns.

Besides libp2p being required for @helia/ipns, what's the reason you'd still want libp2p? If you're only fetching from gateways, you shouldn't even need delegated routing.

achingbrain:
> Besides libp2p being required for @helia/ipns, what's the reason you'd still want libp2p? 

Basically anything that uses peer/content routing, which is most things outside of block retrieval.  The future is for PUTs on gateways so by configuring HTTP peer/content routers we can use the same code paths as everything else and not need to maintain extra libraries.

2color:
> Basically anything that uses peer/content routing, which is most things outside of block retrieval.

The whole point of @helia/http is to be focused solely on verified block retrieval from gateways, which wouldn't use peer/content routing.

> The future is for PUTs on gateways so by configuring HTTP peer/content routers we can use the same code paths as everything else and not need to maintain extra libraries.

What's the source for this? (are you referring to data onboarding via gateways?). This sounds like a premature optimisation for a use-case that hasn't even been planned.

achingbrain:
> The whole point of @helia/http is to be focused solely on verified block retrieval from gateways

If this is the case then maybe it doesn't belong in the `@helia` namespace since this implies a certain amount of compatibility with the other modules in the ecosystem.

I think it's completely reasonable for a user to assume that `@helia/http` will work with `@helia/ipns` for example, which can be enabled by configuring a bare-bones libp2p with [@helia/delegated-routing-v1-http-api-client](https://www.npmjs.com/package/@helia/delegated-routing-v1-http-api-client) to support [IPIP-379](https://github.com/ipfs/specs/pull/379).

I'm interested in the resistance to this, could you speak a bit more to that?

packages/core/README.md
```diff
@@ -0,0 +1,49 @@
...
+  </a>
+</p>
+
+[![ipfs.tech](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](https://ipfs.tech)
+[![Discuss](https://img.shields.io/discourse/https/discuss.ipfs.tech/posts.svg?style=flat-square)](https://discuss.ipfs.tech)
+[![codecov](https://img.shields.io/codecov/c/github/ipfs/helia.svg?style=flat-square)](https://codecov.io/gh/ipfs/helia)
+[![CI](https://img.shields.io/github/actions/workflow/status/ipfs/helia/main.yml?branch=main\&style=flat-square)](https://github.com/ipfs/helia/actions/workflows/main.yml?query=branch%3Amain)
+
+> An implementation of IPFS in JavaScript
+
```
SgtPooki, line 13: We should clarify what the difference between `@helia/core` and `helia` are

packages/http/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+   * webworker), pass true here to hold the gc lock in this process.
+   */
+  holdGcLock?: boolean
+
+  /**
+   * An optional logging component to pass to libp2p. If not specified the
+   * default implementation from libp2p will be used.
+   */
+  logger?: ComponentLogger
+}
```
SgtPooki, line 114: It would be great if we could re-use the config from `@helia/core` or `helia`

packages/http/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+ * Create and return a Helia node
+ */
+export async function createHeliaHTTP (init: HeliaHTTPInit = {}): Promise<Helia> {
+  const datastore = init.datastore ?? new MemoryDatastore()
+  const blockstore = init.blockstore ?? new MemoryBlockstore()
+
+  let libp2p: Libp2p<DefaultLibp2pServices>
+
+  if (isLibp2p(init.libp2p)) {
+    libp2p = init.libp2p as any
```
SgtPooki, line 126:
```suggestion
    libp2p = init.libp2p
```

`isLibp2p` guard clause should set the type appropriately here.

SgtPooki (approved):

packages/http/src/index.ts
achingbrain: Maybe we could extend some base config from `@helia/core`?  Wouldn't want to use the config from `helia` because it'd create a dependency.

achingbrain:
> The boundaries between libp2p and Helia are a bit blurry.

@2color I've opened https://github.com/ipfs/helia/issues/378 - I think it might draw a bit more of a dividing line between them?

> The need for this came up in the context of dapp developers who are currently doing async fetch of unixfs data/json from trusted gateways.

Ah, ok cool - this sounds like the sort of use case we'd be encouraging people to use https://github.com/ipfs/web3-fetch/ for?

Obviously this PR is needed to implement that functionality but it's got the nice familiar fetch-style API, etc.

2color:
> @2color I've opened https://github.com/ipfs/helia/issues/378 - I think it might draw a bit more of a dividing line between them?

Thank you! 

>Ah, ok cool - this sounds like the sort of use case we'd be encouraging people to use https://github.com/ipfs/web3-fetch/ for?

Exactly!

> Obviously this PR is needed to implement that functionality but it's got the nice familiar fetch-style API, etc.

Yep! 💯

packages/http/src/index.ts
achingbrain: `isLibp2p` doesn't guarantee the generic type - e.g. that `init.libp2p.services` satisfies `DefaultLibp2pServices`.

packages/http/src/index.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Create and return a Helia node
+ */
+export async function createHeliaHTTP (init: HeliaHTTPInit = {}): Promise<Helia> {
+  const datastore = init.datastore ?? new MemoryDatastore()
+  const blockstore = init.blockstore ?? new MemoryBlockstore()
+
+  let libp2p: Libp2p<DefaultLibp2pServices>
```
2color, line 123: Will this libp2p instance have all the `DefaultLibp2pServices`? For example, when only used for delegated routing is `dht: KadDHT` one of the services available?

achingbrain: For `@helia/http` `DefaultLibp2pServices` is defined here: https://github.com/ipfs/helia/pull/372/files/283032c685ebf7a4644913ffad9fbe0cc92ea55b#diff-9015afc54277fbde1b6b43d891b88d0ab906d3985fff9e1d560089a1ab2c13e2R5-R7

The only service configured is the delegated routing v1 http api client.  As I said on the call it only does HTTP things, no DHT, no transports, no pubsub, no nothing.

packages/http/src/index.ts
SgtPooki: > Maybe we could extend some base config from @helia/core

I'm up for whichever you think is most maintainable long term. We could update that later if/when we need to abstract further too

packages/core/README.md
achingbrain: Done!

packages/interface/src/index.ts
```diff
@@ -16,22 +16,18 @@
...
 /**
- * The API presented by a Helia node.
+ * The API presented by a Helia node
  */
-export interface Helia<T = Libp2p> {
-  /**
-   * The underlying libp2p node
-   */
-  libp2p: T
-
```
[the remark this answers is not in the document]
achingbrain: Yes, I've updated the PR title

packages/helia/src/index.ts
```diff
@@ -153,11 +159,17 @@ export async function createHelia (init: HeliaInit = {}): Promise<Helia<unknown>
...
-  const helia = new HeliaImpl({
+  const helia = new HeliaP2P({
     ...init,
     libp2p,
     datastore,
-    blockstore
+    blockstore,
+    blockBrokers: init.blockBrokers ?? [
+      bitswap()
+    ],
```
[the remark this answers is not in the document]
achingbrain: Oops, have reinstated it.

packages/core/src/routing.ts
```diff
@@ -0,0 +1,169 @@
...
+      if (peer == null) {
+        continue
+      }
+
+      // deduplicate peers
+      if (seen.has(peer.id)) {
+        continue
+      }
+
+      seen.add(peer.id)
```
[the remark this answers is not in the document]
achingbrain: It's possible but it's hard to see how we can ensure this without first waiting for all routers to finish yielding peers which will be very slow.

That is to say, imagine we have 100 peers yielded by two routers, the first peer from the first router is the same as the last peer from the second router except it has an additional multiaddr - we'd have to wait for both to finish yielding peers, dedupe the list and merge multiaddrs before yielding the first peer.

Perhaps we can revisit if this causes problems in the wild.

packages/http/README.md
```diff
@@ -0,0 +1,70 @@
...
+[![ipfs.tech](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](https://ipfs.tech)
+[![Discuss](https://img.shields.io/discourse/https/discuss.ipfs.tech/posts.svg?style=flat-square)](https://discuss.ipfs.tech)
+[![codecov](https://img.shields.io/codecov/c/github/ipfs/helia.svg?style=flat-square)](https://codecov.io/gh/ipfs/helia)
+[![CI](https://img.shields.io/github/actions/workflow/status/ipfs/helia/main.yml?branch=main\&style=flat-square)](https://github.com/ipfs/helia/actions/workflows/main.yml?query=branch%3Amain)
+
+> A lightweight implementation of IPFS over HTTP in JavaScript
+
+# About
+
+Exports a `createHeliaHTTP` function that returns an object that implements a lightweight version of the Helia API that functions only over HTTP.
```
2color, line 16:
```suggestion
Exports a `createHeliaHTTP` function that returns an object that implements a lightweight version of the Helia API that functions only over HTTP. By default, content and peer routing are requests are resolved using the [Delegated HTTP Routing API](https://specs.ipfs.tech/routing/http-routing-v1/) and blocks are fetched from [Trustless Gateways](https://specs.ipfs.tech/http-gateways/trustless-gateway/). 
```

packages/http/src/index.ts
```diff
@@ -0,0 +1,65 @@
+/**
+ * @packageDocumentation
```
2color, line 2: Is this only used for the API docs generated by typedoc?

packages/interface/src/index.ts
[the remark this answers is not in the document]
2color: Is that the exclamation mark that does a major version bump?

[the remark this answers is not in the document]
achingbrain: Yes, the standard is called [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/)

packages/http/src/index.ts
achingbrain: TypeDoc uses it and we also compile this block to markdown and copy it into the README file

packages/http/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Pass it to other modules like @helia/unixfs to fetch files from the distributed web.
+
+## Example
```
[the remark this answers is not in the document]
achingbrain: Please can you suggest an edit or push a commit to this PR?

packages/routers/README.md
```diff
@@ -0,0 +1,49 @@
...
+    <img src="https://raw.githubusercontent.com/ipfs/helia/main/assets/helia.png" alt="Helia logo" width="300" />
+  </a>
+</p>
+
+[![ipfs.tech](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](https://ipfs.tech)
+[![Discuss](https://img.shields.io/discourse/https/discuss.ipfs.tech/posts.svg?style=flat-square)](https://discuss.ipfs.tech)
+[![codecov](https://img.shields.io/codecov/c/github/ipfs/helia.svg?style=flat-square)](https://codecov.io/gh/ipfs/helia)
+[![CI](https://img.shields.io/github/actions/workflow/status/ipfs/helia/main.yml?branch=main\&style=flat-square)](https://github.com/ipfs/helia/actions/workflows/main.yml?query=branch%3Amain)
+
+> Routers for Helia
```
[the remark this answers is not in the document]
achingbrain: See [this comment](https://github.com/ipfs/helia/pull/372#discussion_r1453389667) about changing the `About` section in the README by updating the `packageDocumentation` block instead.

2color (commented):
Neat having the routers abstraction, thanks for getting this in shape! 

I left some a few cosmetic comments, but looks generally good.

packages/http/README.md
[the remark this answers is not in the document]
2color: Here too, the edit should be to the `index.ts` followed by running npm run docs`, right?

packages/http/src/index.ts
2color: I ran `npm run docs` in the root of the repo and got many typescript errors after this error:

```
Command failed with exit code 3: typedoc . --entryPointStrategy packages --out .docs --plugin /Users/danielnorman/dev/helia/node_modules/aegir/src/docs/unknown-symbol-resolver-plugin.js --plugin /Users/danielnorman/dev/helia/node_modules/aegir/src/docs/type-indexer-plugin.js --plugin typedoc-plugin-mdn-links --plugin /Users/danielnorman/dev/helia/node_modules/aegir/src/docs/readme-updater-plugin.js
[warning] The 'typedoc' key in ./packages/ipns/package.json was used by the legacy-packages entryPointStrategy and will be ignored.
````

**Comment by achingbrain** [comment]
I'm going to merge this so we get RCs published for further experimentation.

**Comment by ptoner** [comment]
Hi,

I'm struggling a little bit to understand these changes. Is it possible to use Helia without any libp2p stuff currently? My use case is that I'm just writing some files to MFS and then creating a CAR file offline which I distribute manually when necessary.

Recently I upgraded to the latest version and suddenly I'm running into some issues with my code.

```
        const blockstore = new FsBlockstore('./data/ipfs/blockstore')
        const datastore = new FsDatastore('./data/ipfs/datastore')
    
        await datastore.open()
        await blockstore.open()
        
        const helia = await createHelia({
          blockstore: blockstore,
          datastore: datastore,
          libp2p: {
            start: false,
            connectionManager: {
              minConnections: 0,
            },
            services: {},
            peerDiscovery: [],
          }
        })
    
        await helia.libp2p.stop()
    
        console.log("IPFS initialized")
```

This is code that was working but now I'm having trouble getting it to build/run. I saw a comment in the init code that suggested moving start: false to the root level of the helia config and this does fix my build issues but now I'm getting an error that says "Service "@libp2p/circuit-relay-v2-transport" required capability "@libp2p/identify" but it was not provided by any component, you may need to add additional configuration when creating your node." when I start.

Is there a different route to accomplish this now?

Appreciate any help, thanks.

PR #374, opened 2024-01-10
fix: use path gateway in default gateways

### LINKED ISSUE #347: Trustless gateway broker should use subdomain gateway to prevent 301 roundtrips
## Background

The current Trustless gateway broker uses path resolution by default and has 4 subdomain gateways defined in the defaults. This leads to it getting 301s for most requests because subdomain gateways redirect from path requests to the subdomain. For example: 

```
https://cf-ipfs.com/ipfs/QmTUAobz6YU1Ej5dTibodQPNtKk2TNnQek5TXHv13Dk2zq?format=raw -> 301 https://bafybeicmg6dvbhclp6sctn7nfhaex64x7fuw552cufodvjt56zispmvhoy.ipfs.cf-ipfs.com/

https://dweb.link/ipfs/QmRCKiUi5g4w4QBpyNkTiczrcTiwQtGK9DHLnncxXdfD7c?format=raw -> 301 https://bafybeibkodyc2gs43vfj4cond2tw2kb4qsuik72bmeij5dpvwiyvsfasl4.ipfs.dweb.link/?format=raw
```

Since 301s are permanent, the browser will cache these after the initial request, however, this results in an additional round trip per CID.

https://github.com/ipfs/helia/blob/299bb0942bbfae492db938c4ccad4e835bab2dbd/packages/helia/src/block-brokers/trustless-gateway/trustless-gateway.ts#L48-L53

## Suggested solutions

- By default, use subdomain gateway URLs to fetch blocks in the gateway broker.
- Use path gateways
- Maybe make the gateway broker aware of whether a gateway supports subdomain resolution, [since some of the public trustless gateways don't support subdomain resolution](https://ipfs.github.io/public-gateway-checker/)

2color:
Fetching from subdomain gateways like cf-ipfs.com using path resolution returns a redirection (same thing with dweb.link) for every CID. This means an additional round trip for every CID requested.

This PR changes the default gateways to their path equivalent to avoid the additional round trip.
fixes #347


packages/block-brokers/src/trustless-gateway/index.ts
```diff
@@ -5,10 +5,10 @@ import type { ProgressEvent } from 'progress-events'
 
 export const DEFAULT_TRUSTLESS_GATEWAYS = [
   // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs-public-gateway-checker.on.fleek.co/
-  'https://dweb.link',
+  'https://ipfs.io',
```
SgtPooki, line 8: We originally discussed not including `ipfs.io` in this list.

packages/block-brokers/src/trustless-gateway/index.ts
```diff
@@ -5,10 +5,10 @@ import type { ProgressEvent } from 'progress-events'
...
 
   // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs-public-gateway-checker.on.fleek.co/
-  'https://cf-ipfs.com',
+  'https://cloudflare-ipfs.com',
```
SgtPooki, line 11: Why switch from cf-ipfs.com to cloudflare-ipfs.com?

SgtPooki (changes requested):

packages/block-brokers/src/trustless-gateway/index.ts
2color: Why is that? Dweb.link and ipfs.io are backed by the same infra.

SgtPooki: I believe the goal is to use ipfs.io gateway less. Plans were to use the new `someguy` / `rainbow` backed gateway.

packages/block-brokers/src/trustless-gateway/index.ts
SgtPooki: Yeah, it makes sense now. Since cloudflare-ipfs.com doesn't support subdomains, it doesn't redirect the path routes to subdomains.

packages/block-brokers/src/trustless-gateway/index.ts
2color: As mentioned above, both ipfs.io and dweb.link are resolving to the same IP (209.94.90.1) and underlying infrastructure. Some of that infrastructure is already using rainbow. 

Are you suggesting that we don't include either one of `dweb.link` or `ipfs.io`?

ns4plabs: We have a dedicated domain `trustless-gateway.link` for the trustless gateway which is backed by [rainbow](https://github.com/ipfs/rainbow)

SgtPooki: trustless-gateway.link should prevent redirects as well:

```sh
curl -H "Accept: application/vnd.ipld.raw" "https://trustless-gateway.link/ipfs/QmTUAobz6YU1Ej5dTibodQPNtKk2TNnQek5TXHv13Dk2zq?format=raw" -vvv > QmTUAobz6YU1Ej5dTibodQPNtKk2TNnQek5TXHv13Dk2zq.json
```

SgtPooki, line 8:
```suggestion
  'https://trustless-gateway.link',
```

SgtPooki (changes requested):

2color:
Blocked until https://github.com/protocol/bifrost-community/issues/4 is resolved.

SgtPooki:
this should be ready to be merged now

PR #381, opened 2024-01-16
feat: expose configured dag walkers and hashers on helia interface

### LINKED ISSUE #375: Exporting a CAR with `@helia/car` of a `@helia/json` encoded CID throws an error
As per @achingbrain recommendation, I tried the `@helia/json` library to try out some simple utf-8 encoded JSON as an alternative to UnixFS.

The APIs are indeed nicer to work with(adding and getting CIDs is an async operation rather than an async iterator), but exporting a CAR fails with the following error:

```
Uncaught (in promise) Error: No dag walker found for cid codec 512
    at #walkDag (index.js:129:19)
    at eval (index.js:109:36)
    at eval (index.js:187:37)
    at #tryToStartAnother (index.js:125:17)
    at eval (index.js:211:36)
    at new Promise (<anonymous>)
    at PQueue.add (index.js:181:16)
    at DefaultCar.export (index.js:108:24)
    at createCarBlob (ipfs.tsx:118:20)
    at async eval (sign-page.tsx:51:18)
```

The source that produces this looks as follows

```typescript
import { json as heliaJson } from '@helia/json'

export async function createCidForSignedMessage(
  helia: Helia,
  { message, signature, address }: SignedMessagePayload,
) {
  const j = heliaJson(helia)

  const cid = await j.add({
    address,
    message,
    signature,
  })

  return cid
}

export async function createCarBlob(helia: Helia, cid: CID) {
  const { writer, out } = await CarWriter.create(cid)
  car(helia).export(cid, writer)

  const carBlob = await carWriterOutToBlob(out)
  return carBlob
}
```

---

I tested this using `@helia/dag-json` and this works. So for now that's what I'm using since it gives the same return type.

achingbrain:
To allow modules like `@helia/car` to re-use configured dag walkers instead of having to bundle their own, expose configured dag walkers and hashers on the helia interface and have them consume those.

Fixes #375


achingbrain:
Will merge to main after https://github.com/ipfs/helia/issues/375 has gone in.

PR #384, opened 2024-01-16
feat: export binary from @helia/interop

achingbrain:
To prevent things like transient deps breaking Kubo interop tests,
export a bin to allow installing the interop tests from npm and
running them in an easy-to-digest way.


packages/interop/src/bin.ts
```diff
@@ -0,0 +1,18 @@
+#! /usr/bin/env node
+/* eslint-disable no-console */
+
+import { spawn } from 'node:child_process'
+
+const test = spawn('npx', ['aegir', 'test'])
```
SgtPooki, line 6: I don't think we need `npx` here but not sure of the hassle of getting aegir binary

SgtPooki (approved):

packages/interop/src/bin.ts
achingbrain: `aegir` is installed as a dep - we need to find the binary so we could assume the location with:

```js
spawn('node', ['node_modules', '.bin', 'aegir', 'test'])
```

or use `npx` and let npm figure it out.

PR #391, opened 2024-01-17
fix: publish shrinkwrap file with @helia/interop

achingbrain:
To shield consumers from transient dep updates breaking interop tests between releases, publish with a shrinkwrap file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules
 package-lock.json
 yarn.lock
 .vscode
+npm-shrinkwrap.json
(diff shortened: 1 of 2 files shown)


SgtPooki (approved):

achingbrain:
I'm not 100% convinced this is a good idea.  For example if we fix a bug in a dependency, it would be blocked on a release here before anyone would see the fix.

This would become problematic if we were experimenting with an unrelated RC, for example.

achingbrain:
Closing as consumers should use lockfiles if they desire complete stability.

Issue #348, opened 2023-12-14
feat: Helia fetch for dApps


SgtPooki:
this task is intended to cover all of the actions necessary to get a library, lets call it "helia fetch" working for dApps.

- [ ] Create library that accepts HTTP Request and returns HTTP Responses
- [ ] do unix-fs-exporter things (sub resource loading)
- [ ] bring in learnings from https://github.com/ipfs-shipyard/helia-service-worker-gateway/blob/main/src/lib/heliaFetch.ts & https://github.com/ipfs/helia-http-gateway

ping ed from liquity to see what he's working on

# Ecosystem Needs

(lidel) One way to look at the body of work ahead is to identify needs of different stakeholder, and focus on delivering useful software for the wider audience first. Below if my first stab in surfacing needs and related scope of work, feel free to adjust as we go.

## Food and Shelter for Web developers and public gateways

Use cases: 

- web dev new to content-addressing: wants to "add IPFS to their website" and load image/video by CID, but does not want to spend hours learning about our stack. they are familiar with how web works, are comfortable with Fetch API and Service Worker
- website owner: being able to offload and verify fetch og content-addressed subresources without depending on a single gateway
- ipfs.io: we want to enable public good gateways to start throttling websites which hotlink images and videos, and return an error message which encourages leeching websites to switch to our helia-fetch library (mutual benefit, which removes their dependency on a single gateway).

Needs:

- [ ] Turn-key NPM package developers can include in their projects
- [ ] API modelled after [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch): HTTP Request → HTTP Response
  - Prior art: https://github.com/RangerMauve/js-ipfs-fetch
  - Good abstraction, can be reused in both Web Apps and HTTP gateway implementations
  - UX: suppport  both `ipfs://` and `/ipfs/` paths
- [ ] Support codecs and hashes of defined in [Web Pathing Spec Draft](https://github.com/ipfs/specs/pull/453/files) (dag-pb/raw/dag-cbor/dag-json, UnixFS HAMTs)
- [ ] Minimal subset of [path gateways](https://specs.ipfs.tech/http-gateways/path-gateway/) required for hosting of subresource (file) hosting.
  - No need for `_redirects`, dir listing, or CAR responses.
  - Range requests are required.
  - (optional) IPNS can be added later (focus on immutables first)
- [ ] Ability to configure source(s) of blocks
[...]

PR #392, opened 2024-01-17
feat: create @helia/verified-fetch

SgtPooki:
Fixes https://github.com/ipfs/helia/issues/348

```[tasklist]
- [x] Implement OnProgress options property
- [x] add tests for verifiedFetch instance (stubbing ipns/unixfs and verifying that calls are made appropriately)
- [x] disable delegated routing in browser due to CORS issue at https://github.com/ipshipyard/waterworks-community/issues/4
- [x] Ensure onProgress events emitted are agreed upon
- [x] check CID.code type and use the appropriate library to get the resultant data
- [x] confirm e2e/interop tests for use-cases we want covered.
- [x] 100% (or close to) coverage at https://app.codecov.io/github/ipfs/helia/tree/feat%2Fcreate-verified-fetch/packages%2Fverified-fetch%2Fsrc
- [x] cache ipns results in verified fetch (ipfs.tech results in 16k dns requests in service-worker-gateway....) for at least a minute
- [x] add error 501 returns for non-supported codecs for now (users will file issues when they need support)
```


packages/verified-fetch/package.json
```diff
@@ -0,0 +1,100 @@
...
+    "@helia/ipns": "^4.0.0",
+    "@types/mime-types": "^2.1.4",
+    "@types/sinon": "^17.0.2",
+    "aegir": "^42.1.0",
+    "blockstore-core": "^4.3.10",
+    "datastore-core": "^9.2.7",
+    "delay": "^6.0.0",
+    "it-all": "^3.0.4",
+    "sinon": "^17.0.1",
+    "sinon-ts": "^2.0.0"
```
SgtPooki, line 94: these need cleaned up, as much was copied over from @helia/utils

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,101 @@
...
+ * @packageDocumentation
+ *
+ * Exports a `createVerifiedFetch` function that returns a `fetch()` like API method {@link Helia} for fetching IPFS content.
+ *
+ * You may use any supported resource argument to fetch content:
+ *
+ * - CID string
+ * - CID instance
+ * - IPFS URL
+ * - IPNS URL
```
SgtPooki, line 11: I think we should remove CID string and CID instance from possible resource arguments.

packages/verified-fetch/src/interface.ts
```diff
@@ -0,0 +1,22 @@
+import type { CID } from 'multiformats/cid'
+
+/**
+ * Instead of passing a Helia instance, you can pass a list of gateways and routers, and a Helia instance will be created for you.
```
SgtPooki, line 4:
```suggestion
 * Instead of passing a Helia instance, you can pass a list of gateways and routers, and a HeliaHTTP instance will be created for you.
```

packages/verified-fetch/src/utils/get-content-type.ts
```diff
@@ -0,0 +1,55 @@
...
+import mime from 'mime-types'
+
+interface testInput {
+  bytes: Uint8Array
+  path: string
+}
+
+type testOutput = Promise<string | undefined>
+
+export const DEFAULT_MIME_TYPE = 'text/html'
```
SgtPooki, line 12: we should decide on what mime-type makes sense if we cannot determine one.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  // handle vnd.ipfs.ipns-record
+  private async handleIPNSRecord ({cid, path, options}: {cid: CID, path: string, options?: VerifiedFetchOptions}): Promise<Response> {
+    return new Response('TODO: handleIPNSRecord', { status: 500 })
+  }
+
+  // handle vnd.ipld.car
+  private async handleIPLDCar ({cid, path, options}: {cid: CID, path: string, options?: VerifiedFetchOptions}): Promise<Response> {
+    return new Response('TODO: handleIPLDCar', { status: 500 })
+  }
```
SgtPooki, line 108: I don't believe we can use @helia/http for fetching this content the way @helia/block-brokers is currently set up. We will need to fetch from gateways directly for these without re-architecting @helia/http or block-brokers.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,153 @@
...
+    if (response == null) {
+      response = await this.handleIPLDRaw({cid, path, options})
+    }
+
+    response.headers.set('etag', cid.toString())
+    // response.headers.set('cache-cotrol', 'public, max-age=29030400, immutable')
+    response.headers.set('cache-cotrol', 'no-cache') // disable caching when debugging
+    response.headers.set('x-ipfs-path', path)
+    response.headers.set('x-ipfs-cid', cid.toString())
+    response.headers.set('x-ipfs-protocol', 'ipfs')
```
SgtPooki, line 149: headers set here so we can share them across methods. unique headers based on content-type should be set in appropriate methods

packages/verified-fetch/README.md
```diff
@@ -0,0 +1,204 @@
...
+[![ipfs.tech](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](https://ipfs.tech)
+[![Discuss](https://img.shields.io/discourse/https/discuss.ipfs.tech/posts.svg?style=flat-square)](https://discuss.ipfs.tech)
+[![codecov](https://img.shields.io/codecov/c/github/ipfs/helia.svg?style=flat-square)](https://codecov.io/gh/ipfs/helia)
+[![CI](https://img.shields.io/github/actions/workflow/status/ipfs/helia/main.yml?branch=main\&style=flat-square)](https://github.com/ipfs/helia/actions/workflows/main.yml?query=branch%3Amain)
+
+> A fetch-like API for IPFS content on the web.
+
+# About
+
+`@helia/verified-fetch`` is a library that provides a fetch-like API for fetching content from IPFS. This library should act as a replacement for the `fetch()` API for fetching content from IPFS, and will return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that can be used in a similar manner to the `fetch()` API. This means browser and HTTP caching inside browser main threads, web-workers, and service workers, as well as other features of the `fetch()` API should work in a way familiar to developers.
```
2color, line 16:
```suggestion
`@helia/verified-fetch` is a library that provides a fetch-like API for fetching content from IPFS. This library should act as a replacement for the `fetch()` API for fetching content from IPFS, and will return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that can be used in a similar manner to the `fetch()` API. This means browser and HTTP caching inside browser main threads, web-workers, and service workers, as well as other features of the `fetch()` API should work in a way familiar to developers.
```

packages/verified-fetch/src/utils/get-content-type.ts
lidel: application/octet-stream

packages/verified-fetch/README.md
```diff
@@ -0,0 +1,204 @@
...
+You can see variations of Helia and js-libp2p configuration options at https://helia.io/interfaces/helia.index.HeliaInit.html.
+
+The `@helia/http` module is currently in-progress, but the init options should be a subset of the `helia` module's init options. See https://github.com/ipfs/helia/issues/289 for more information.
+
+```ts
+import { createVerifiedFetch } from '@helia/verified-fetch'
+import { CreateHelia as CreateHeliaHttpOnly } from '@helia/http'
+
+const verifiedFetch = await createVerifiedFetch(
+  CreateHeliaHttpOnly({
```
achingbrain, line 59:
```suggestion
import { createHelia as createHeliaHttpOnly } from '@helia/http'

const verifiedFetch = await createVerifiedFetch(
  createHeliaHttpOnly({
```

packages/verified-fetch/src/index.ts
lidel: Asking users who already have these to do extra work of prepending `ipfs://` feels like a papercut.
Is there harm in keeping them? Are we worried about typescript types breaking constantly?

packages/verified-fetch/src/verified-fetch.ts
lidel: Hm.. do we need to invent `x-ipfs-cid` and `x-ipfs-path`? 
[`X-Ipfs-Path`](https://specs.ipfs.tech/http-gateways/path-gateway/#x-ipfs-path-response-header) already includes protocol (ipfs or ipns), and if we want CIDs for sub-paths, there is [`X-Ipfs-Roots`](https://specs.ipfs.tech/http-gateways/path-gateway/#x-ipfs-roots-response-header)

packages/verified-fetch/src/verified-fetch.ts
aschmahmann: 1. My suspicion is that we can do this can be done with helia/http by taking the blocks and assembling them into CARs and validating them
2. I don't suspect anyone will need support for this (at least in a browser environment) for quite a while. For now I'd just reject requests for `vnd.ipld.car` as unsupported

2color:
As discussed, for a long (multiblock) UnixFS JSON, try https://cloudflare-ipfs.com/ipns/tokens.uniswap.org which resolves to `QmQJ8fxavY54CUsxMSx9aE9Rdcmvhx8awJK2jzJp4iAqCr`

packages/verified-fetch/README.md
```diff
@@ -0,0 +1,204 @@
...
+
+You can see variations of Helia and js-libp2p configuration options at https://helia.io/interfaces/helia.index.HeliaInit.html.
+
+The `@helia/http` module is currently in-progress, but the init options should be a subset of the `helia` module's init options. See https://github.com/ipfs/helia/issues/289 for more information.
+
+```ts
+import { createVerifiedFetch } from '@helia/verified-fetch'
+import { CreateHelia as CreateHeliaHttpOnly } from '@helia/http'
+
+const verifiedFetch = await createVerifiedFetch(
```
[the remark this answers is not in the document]
SgtPooki: I've got that in my test but didn't update the readme yet. will do

packages/verified-fetch/src/index.ts
SgtPooki: The concern is the additional "path" argument that is required if a user provides a CID instance. We discussed during helia WG. I will remove `CID string` as an option and user who pass CID instances won't be able to provide a path

SgtPooki:
> As discussed, for a long (multiblock) UnixFS JSON, try [cloudflare-ipfs.com/ipns/tokens.uniswap.org](https://cloudflare-ipfs.com/ipns/tokens.uniswap.org) which resolves to QmQJ8fxavY54CUsxMSx9aE9Rdcmvhx8awJK2jzJp4iAqCr

@2color a test for this has been added and is working properly.

packages/verified-fetch/test/gateways.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    const {cid} = await addContentToKuboNode(controller, content.marshal())
+    expect(cid).to.be.ok()
+    // @ts-expect-error - todo fix types
+    const resp = await verifiedFetch(cid)
+    expect(resp).to.be.ok()
+    const text = await resp.text() // this currently has UnixFS data in it, and should not when returned from verified-fetch
+
+    // the below commented lines will get the test to pass, but we need to move this into verified fetch
+    // const marshalledResponseData = await resp.arrayBuffer()
+    // const encodedText = UnixFS.unmarshal(new Uint8Array(marshalledResponseData)).data
```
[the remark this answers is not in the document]
SgtPooki: I'm doing this via https://github.com/ipfs/helia/pull/392/files#diff-12157866119b95f1b476d4e54ac9d3f22a24cc7e7af1a85f251c00c4dfbefcc6R135 now, but it's not sophisticated at all. I'm not sure when unixfs data for multipart content needs unmarshalled though.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * handle vnd.ipld.raw
+   * This is the default method for fetched content.
+   */
+  private async handleIPLDRaw ({ cid, path, options }: { cid: CID, path: string, options?: VerifiedFetchOptions }): Promise<Response> {
+    // const finalFileStat = await this.unixfs.stat(cid, { path, signal: options?.signal })
+
+    const asyncIter = this.unixfs.cat(cid, { path, signal: options?.signal })
```
aschmahmann, line 131: Using `cat` here doesn't seem like  you want since you're returning the raw block. Instead you probably want:

https://github.com/ipfs/helia/blob/031519cea8e45a82272d3d56c7fd44f481184989/packages/unixfs/src/commands/utils/resolve.ts#L35

Followed by loading the block corresponding to the returned CID

SgtPooki: that method isn't exported, so i've switched to `.stat` for now

SgtPooki:
FYI tests are failing when ran locally with `Failed to load resource: net::ERR_CONNECTION_REFUSED`. I need to find a better way to run a gateway that can serve this content, or a better way around writing tests to verify that we're getting the content we expect back.

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: opened https://github.com/ipfs/helia/issues/399 for exporting that method, since we would need that method (or similar) for `X-Ipfs-Roots`

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,187 @@
...
+  constructor ({ helia, ipns, unixfs, dagJson, json }: VerifiedFetchConstructorOptions) {
+    this.helia = helia
+    this.ipns = ipns ?? heliaIpns(helia, {
+      resolvers: [
+        dnsJsonOverHttps('https://mozilla.cloudflare-dns.com/dns-query'),
+        dnsOverHttps('https://mozilla.cloudflare-dns.com/dns-query'),
+        dnsOverHttps('https://cloudflare-dns.com/dns-query'),
+        dnsOverHttps('https://dns.google/dns-query'),
+        dnsJsonOverHttps('https://dns.google/resolve'),
+        dnsOverHttps('https://dns.quad9.net/dns-query')
```
aschmahmann, line 38: It doesn't look like https://dns.quad9.net supports cors 
image

packages/verified-fetch/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+1. `onProgress` - Similar to Helia `onProgress` options, this will be a function that will be called with a progress event. Supported progress events are:
+    - `helia:verified-fetch:error` - An error occurred during the request.
+    - `helia:verified-fetch:request:start` - The request has been sent
+    - `helia:verified-fetch:request:complete` - The request has been sent
+    - `helia:verified-fetch:request:error` - An error occurred during the request.
+    - `helia:verified-fetch:request:abort` - The request was aborted prior to completion.
+    - `helia:verified-fetch:response:start` - The initial HTTP Response headers have been set, and response stream is started.
+    - `helia:verified-fetch:response:complete` - The response stream has completed.
+    - `helia:verified-fetch:response:error` - An error occurred while building the response.
```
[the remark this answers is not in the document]
SgtPooki: onProgress events should help answer the following questions:

* did I finish resolving IPNS name / peerId
* was I able to find the terminal element, if not, why not.
* were there any redirects?
* download progress for terminal elements (chunk counter?)
* did we find who has the data (providers)?

packages/verified-fetch/src/interface.ts
```diff
@@ -0,0 +1,22 @@
...
+ * The types for the first argument of the `verifiedFetch` function.
+ */
+export type ResourceType = string | CID
+
+/**
+ * The second argument of the `verifiedFetch` function.
+ */
+export interface VerifiedFetchOptions extends RequestInit {
+  signal?: AbortSignal
+  // onProgress?: (progress: any) => void
```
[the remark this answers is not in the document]
SgtPooki: see https://github.com/ipfs/helia/pull/392#discussion_r1466502990 for discussion around events to impl

packages/verified-fetch/README.md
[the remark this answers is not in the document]
SgtPooki: 'if async custom work, should probably have async event':
custom onprogress:
* redirects?
* progress?

the rest should be piped up from other helia Modules used.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,174 @@
...
+  private async handleIPLDCar ({cid, path, options}: {cid: CID, path: string, options?: VerifiedFetchOptions}): Promise<Response> {
+    return new Response('TODO: handleIPLDCar', { status: 500 })
+  }
+
+  /**
+   * handle vnd.ipld.raw
+   * This is the default method for fetched content.
+   */
+  private async handleIPLDRaw ({cid, path, options}: {cid: CID, path: string, options?: VerifiedFetchOptions}): Promise<Response> {
+    const asyncIter = await this.unixfs.cat(cid, { path, signal: options?.signal })
```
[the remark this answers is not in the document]
SgtPooki: This method is called/named incorrectly: ipldRaw doesn't mean unixfs

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+export class VerifiedFetch {
+  private readonly helia: Helia
+  private readonly ipns: IPNS
+  private readonly unixfs: HeliaUnixFs
+  private readonly dagJson: DAGJSON
+  private readonly json: HeliaJSON
+
+  constructor ({ helia, ipns, unixfs, dagJson, json }: VerifiedFetchConstructorOptions) {
+    this.helia = helia
```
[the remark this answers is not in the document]
SgtPooki: we do, in src/index.ts. This class should not be consumed publicly

SgtPooki:
@lidel I did confirm that loading paths for HAMT-sharded directories does work with [helia-http-gateway using verified-fetch](https://github.com/ipfs/helia-http-gateway/pull/63)

image

However, running a test locally against that hamt-sharded directory seem to take too long. it would be nice if we had a smaller car to use to verify for test environments. 

The local run of https://github.com/ipfs/helia/pull/403 also fails on my machine (it's a 64GB memory M1 Max)

packages/interop/.aegir.js
```diff
@@ -1,12 +1,58 @@
...
 import * as kuboRpcClient from 'kubo-rpc-client'
+import { dirname, join } from 'node:path'
+import { createReadStream } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+import * as http from "node:http";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+// create an http server that will host the fixture data files. When receiving a request for a fileName, it will return './src/fixtures/data/${fileName}'
+async function createFixtureServer() {
```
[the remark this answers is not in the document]
achingbrain: Aegir does this for you - for an example see the use of [loadFixture](https://github.com/ipfs/js-ipfs/blob/bf1bc8b18d750d5591bdd575a6d0f4c067d61d0a/packages/interface-ipfs-core/src/utils/index.js#L3)

packages/verified-fetch/README.md
```diff
@@ -0,0 +1,211 @@
...
+[![ipfs.tech](https://img.shields.io/badge/project-IPFS-blue.svg?style=flat-square)](https://ipfs.tech)
+[![Discuss](https://img.shields.io/discourse/https/discuss.ipfs.tech/posts.svg?style=flat-square)](https://discuss.ipfs.tech)
+[![codecov](https://img.shields.io/codecov/c/github/ipfs/helia.svg?style=flat-square)](https://codecov.io/gh/ipfs/helia)
+[![CI](https://img.shields.io/github/actions/workflow/status/ipfs/helia/main.yml?branch=main\&style=flat-square)](https://github.com/ipfs/helia/actions/workflows/main.yml?query=branch%3Amain)
+
+> A fetch-like API for IPFS content on the web.
+
+# About
+
+`@helia/verified-fetch` is a library that provides a fetch-like API for fetching content from IPFS. This library should act as a replacement for the `fetch()` API for fetching content from IPFS, and will return a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that can be used in a similar manner to the `fetch()` API. This means browser and HTTP caching inside browser main threads, web-workers, and service workers, as well as other features of the `fetch()` API should work in a way familiar to developers.
```
achingbrain, line 16: Everything from `# About` to the next L1 header (`# Install` here) needs to be in the `@packageDocumentation` block of `src/index.ts` as the contents of the readme will be overwritten by this documentation block when `npm run docs` is run.

packages/verified-fetch/.aegir.js
```diff
@@ -0,0 +1,3 @@
+/** @type {import('aegir').PartialOptions} */
+export default {
```
achingbrain, line 2: This config file seem superfluous?

packages/verified-fetch/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Browser `<script>` tag
+
+Loading this module through a script tag will make it's exports available as `HeliaVerifiedFetch` in the global namespace.
+
+```html
+<script src="https://unpkg.com/@helia/verified-fetch/dist/index.min.js"></script>
+```
+
+### Configuration
```
achingbrain, line 46: All of this exposition needs to move into the `@packageDocumentation` block of `/src/index.ts`

packages/verified-fetch/package.json
```diff
@@ -0,0 +1,85 @@
+{
+  "name": "@helia/verified-fetch",
+  "version": "0.0.0",
+  "description": "A fetch-like API for IPFS content on the web.",
```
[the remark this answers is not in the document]
achingbrain: It'd probably be good to get the word "verified" or "trustless" in there?

packages/verified-fetch/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "provenance": true
+  },
+  "keywords": [
+    "IPFS",
+    "fetch",
+    "helia"
+  ],
+  "type": "module",
+  "types": "./dist/src/index.d.ts",
+  "typesVersions": {
```
achingbrain, line 25: No need for `typesVersions` when there's only a single export.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,211 @@
...
+import { type CID } from 'multiformats/cid'
+import { CustomProgressEvent } from 'progress-events'
+import { getStreamAndContentType } from './utils/get-stream-and-content-type.js'
+import { parseResource } from './utils/parse-resource.js'
+import type { CidDetail, ResourceType, VerifiedFetchOptions } from './interface.js'
+import type { Helia } from '@helia/interface'
+
+const log = logger('helia:verified-fetch')
+
+interface VerifiedFetchConstructorOptions {
```
achingbrain, line 16: For consistency, every other class in the monorepo has a split between components and init options in their constructor args.

If you like, components are things you use to get work done, init options are parameters that change how the work is done.

packages/verified-fetch/src/interface.ts
```diff
@@ -0,0 +1,53 @@
+import type { ResolveDnsLinkProgressEvents, ResolveProgressEvents } from '@helia/ipns'
```
achingbrain, line 1: I think most (all?) of these interfaces should be declared in the `index.ts` file.

As implemented they will not be exported from the module, so (for example) the exposed fetch factory function will accept an argument interface that cannot be referenced by consuming code.

This will work due to type overlap but it makes it harder to use and also means the types become undocumented in the generated API docs.

packages/verified-fetch/src/interface.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export type ResourceType = string | CID
+
+export type BubbledProgressEvents =
+  // unixfs
+  GetEvents |
+  // ipns
+  ResolveProgressEvents | ResolveDnsLinkProgressEvents
+
+export interface CidDetail {
```
achingbrain, line 26: CID is an acronym which we keep in capitals elsewhere:

```suggestion
export interface CIDDetail {
```

packages/verified-fetch/src/interface.ts
```diff
@@ -0,0 +1,53 @@
...
+  cid: string
+  path: string
+}
+
+export interface CidDetailError extends CidDetail {
```
achingbrain, line 31: CID is kept in capitals elsewhere:

```suggestion
export interface CIDDetailError extends CIDDetail {
```

packages/verified-fetch/src/interface.ts
```diff
@@ -0,0 +1,53 @@
...
+  ProgressEvent<'verified-fetch:request:start', CidDetail> |
+  ProgressEvent<'verified-fetch:request:info', string> |
+  ProgressEvent<'verified-fetch:request:progress:chunk', CidDetail> |
+  ProgressEvent<'verified-fetch:request:end', CidDetail> |
+  ProgressEvent<'verified-fetch:request:error', CidDetailError>
+
+/**
+ * The second argument of the `verifiedFetch` function.
+ */
+export interface VerifiedFetchOptions extends Omit<RequestInit, 'signal'>, AbortOptions, ProgressOptions<BubbledProgressEvents | VerifiedFetchProgressEvents> {
```
achingbrain, line 45: Any reason not to just take the `signal` option from `RequestInit` instead of `Omit`ing it, then adding it again with `AbortOptions`?

```suggestion
export interface VerifiedFetchOptions extends RequestInit, ProgressOptions<BubbledProgressEvents | VerifiedFetchProgressEvents> {
```

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (protocol === 'ipfs') {
+    try {
+      cid = CID.parse(cidOrPeerIdOrDnsLink)
+    } catch (err) {
+      log.error(err)
+      throw new TypeError('Invalid CID for ipfs://<cid> URL')
+    }
+  } else {
+    // protocol is ipns
+    if (cidOrPeerIdOrDnsLink.includes('.')) {
```
achingbrain, line 36: Usually we do this the other way round - try to resolve something as a domain last since it's the loosest format.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,211 @@
...
+      } else if (query.format === 'json' || format.includes('application/json')) {
+        return new Response('application/json support is not implemented', { status: 501 })
+      } else if (query.format === 'cbor' || format.includes('application/cbor')) {
+        return new Response('application/cbor support is not implemented', { status: 501 })
+      }
+    }
+
+    if (response == null) {
+      switch (cid.code) {
+        case 0x200:
```
achingbrain, line 175: Can you use the `code` property from the modules in question or some named constants?  We get these numbers wrong sometimes..

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+  start: InstanceType<typeof VerifiedFetch>['start']
+  stop: InstanceType<typeof VerifiedFetch>['stop']
+}
+
+/**
+ * Create and return a Helia node
+ */
+export async function createVerifiedFetch (init: Helia | CreateVerifiedFetchWithOptions): Promise<VerifiedFetchMethod> {
+  let heliaInstance: null | Helia = null
+  if ((init as CreateVerifiedFetchWithOptions).gateways == null) {
```
achingbrain, line 68: You can disambiguate the `init` arg with something like:

```ts
function isHelia (obj?: any): obj is Helia {
  // test for the presence of known Helia properties
  return obj?.blockstore != null && obj?.datastore != null && ...
}
```

...then you can remove the casting.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,211 @@
...
+      onProgress: options?.onProgress
+    })
+    options?.onProgress?.(new CustomProgressEvent<CidDetail>('verified-fetch:request:end', { detail: { cid: cid.toString(), path } }))
+    const response = new Response(JSON.stringify(result), { status: 200 })
+    response.headers.set('content-type', 'application/json')
+    return response
+  }
+
+  /**
+   * handle vnd.ipld.raw
```
achingbrain, line 89: `vnd.ipld.raw` is a block with opaque bytes, this method seems to treat the DAG as if it's something UnixFS-y?

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,94 @@
...
+ * import { createVerifiedFetch } from '@helia/verified-fetch'
+ * import { CID } from 'multiformats/cid'
+ *
+ * const verifiedFetch = await createVerifiedFetch({
+ *  gateways: ['http://mygateway.info', 'http://trustless-gateway.link']
+ * })
+ *
+ * const cid = CID.parse('bafyFoo') // some image file
+ * const response = await verifiedFetch(cid)
+ * const json = await response.json()
```
achingbrain, line 24: Do you think we should call the variable in the example `fetch` to really hammer the point home about API familiarity?

```suggestion
 * const fetch = await createVerifiedFetch({
 *  gateways: ['http://mygateway.info', 'http://trustless-gateway.link']
 * })
 *
 * const cid = CID.parse('bafyFoo') // some image file
 * const response = await fetch(cid)
 * const json = await response.json()
```

achingbrain (changes requested):
LGTM - there's bits of tidy up to be done internally but the API looks solid enough for an RC.

Comments are inline.

packages/verified-fetch/src/index.ts
SgtPooki: I'm good with that, but we're shadowing the global fetch at this point which could cause confusion. Will wait for others' opinions before making this switch.

packages/verified-fetch/src/index.ts
SgtPooki: typeguard ftw

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: 100% good call

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: can do

packages/verified-fetch/src/interface.ts
SgtPooki: I wanted to support future cases where AbortOptions changes, since most Helia methods, and Helia module's methods, take an object with an interface extending AbortOptions. With that said, this module is a little bit different than others, so I think it would be fine if we left with RequestInit

packages/verified-fetch/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "@types/sinon": "^17.0.2",
+    "aegir": "^42.1.0",
+    "helia": "next",
+    "sinon": "^17.0.1",
+    "sinon-ts": "^2.0.0"
+  },
+  "browser": {
+    "node:buffer": false,
+    "node:stream": false
+  }
```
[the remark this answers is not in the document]
SgtPooki: no, this is left-over from when I was trying to use `file-type` for mime-types. will remove

packages/verified-fetch/.aegir.js
SgtPooki: will remove. this was still left from before I moved tests to `interop`

packages/verified-fetch/README.md
SgtPooki: I wasn't sure exactly how that would work. Thanks, will update.

packages/interop/.aegir.js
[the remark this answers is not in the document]
SgtPooki: it would be nice if aegir provided a stream so it wasn't sync. Since for this use-case, in the browser it needs to be fetched from the browser, loaded into kubo-rpc-client, then passed back to kubo node. Would be great if it was streamed all the way through.

packages/verified-fetch/src/utils/parse-url-string.ts
SgtPooki: [`a34119b` (#392)](https://github.com/ipfs/helia/pull/392/commits/a34119b5066f66d18597bed3dd98f8e8d36e3f06)

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,186 @@
...
+      // onProgress: (evt) => {
+      //   log.trace('%s progress event for %c/%s', evt.type, cid, path)
+      // }
+    })
+    if (stat.type === 'directory') {
+      const dirCid = stat.cid
+      // check for redirects
+
+      log.trace('found directory at %c/%s, looking for root files', cid, path)
+      for (const rootFilePath of ['index.html', 'index.htm', 'index.shtml']) {
```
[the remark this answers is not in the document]
SgtPooki: I don't think we're providing any SSI from gateways, so it would be up to a specific service to implement those, but it's worth thinking about what we want to consider root files.

I think the root list should be specified in specs.ipfs.tech if different from typical website loading, and not configurable.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  constructor ({ helia, ipns, unixfs, dagJson, json }: VerifiedFetchConstructorOptions) {
+    this.helia = helia
+    this.ipns = ipns ?? heliaIpns(helia, {
+      resolvers: [
+        dnsJsonOverHttps('https://mozilla.cloudflare-dns.com/dns-query'),
+        dnsOverHttps('https://mozilla.cloudflare-dns.com/dns-query'),
+        dnsOverHttps('https://cloudflare-dns.com/dns-query'),
+        dnsOverHttps('https://dns.google/dns-query'),
+        dnsJsonOverHttps('https://dns.google/resolve')
```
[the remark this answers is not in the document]
lidel: Sgtm, since we only talk to Cloudflare and Google anyway, we can talk JSON with them. If someone needs binary DNS, or ENS, they can pass custom resolver they trust (for ENS DoH at https://resolver.cloudflare-eth.com/dns-query or implement something with https://github.com/a16z/helios)

packages/verified-fetch/src/verified-fetch.ts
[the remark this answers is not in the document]
lidel: Since https://specs.ipfs.tech/http-gateways/path-gateway/#response-payload only looks for `index.html`, and it been this way for newarly a decade, no need to check for anything else.

packages/interop/src/verified-fetch-json.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+        routers: process.env.RUNNER_ENV === 'node' ? [`http://${controller.api.gatewayHost}:${controller.api.gatewayPort}`] : []
+      })
+    })
+
+    after(async () => {
+      await controller.stop()
+      await verifiedFetch.stop()
+    })
+
+    it('handles uniswap tokens list json', async () => {
```
lidel, line 34:
```suggestion
    it('.json() supports UnixFS-chunked JSON file', async () => {
```

SgtPooki:
latest helia/verified-fetch is not resolving some of the dApps Ed provided in the dApps working group (https://hackmd.io/LS4eUSZXTriQ0j8UIjDl4A?view#Agenda)

❌ `/ipns/app.aave.com` is failing ipns resolution even though we are getting an answer back:

image

✅ /ipns/app.spark.fi
❌ /ipns/app.olympusdao.finance

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+      expect(unixfsStub.stat.withArgs(testCID, { path: 'index.html', signal, onProgress: anyOnProgressMatcher }).callCount).to.equal(1)
+      expect(unixfsStub.stat.withArgs(testCID, { path: 'index.htm', signal, onProgress: anyOnProgressMatcher }).callCount).to.equal(1)
+      expect(unixfsStub.stat.withArgs(testCID, { path: 'index.shtml', signal, onProgress: anyOnProgressMatcher }).callCount).to.equal(1)
+      expect(unixfsStub.cat.withArgs(testCID).callCount).to.equal(0)
+      expect(onProgress.callCount).to.equal(0)
+      expect(resp).to.be.ok()
+      expect(resp.status).to.equal(501)
+    })
+
+    it('should return dag-json encoded CID', async () => {
```
lidel, line 231: nit: while we at it, can we add the same `.json()` read test for `dag-cbor` too? CID can be produced by bellow command

```console
$ echo -n '{"hello":"world"}' | ipfs dag put --store-codec dag-cbor
bafyreidykglsfhoixmivffc5uwhcgshx4j465xwqntbmu43nb2dzqwfvae
```

packages/verified-fetch/README.md
lidel, line 1: For the initial release, it would be useful to demonstrate value added by the library by having examples similar to https://github.com/ipfs/helia/tree/main/packages/ipns#example---getting-started right in the README.

We already have tests for these, just need to add HTML+JS snippets to README:


- `dag-cbor` read via `.json()`
- UnixFS JSON read via `.json()`
- plain text file in UnixFS HAMT read via `.text()`
- `` having `src` set to `data:image/jpg;base64,[...]` with payload read via `.arrayBuffer()`

packages/verified-fetch/test/verified-fetch.spec.ts
SgtPooki: added: [`0e0c287` (#392)](https://github.com/ipfs/helia/pull/392/commits/0e0c2870b54cb0fe41e19738abf5e30fc50987e1)

packages/verified-fetch/README.md
SgtPooki: the examples are in @packageDocumentation of src/index.ts now and will be copied over to README when merged

SgtPooki:
> /ipns/app.aave.com is failing ipns resolution even though we are getting an answer back:

I created a demo for testing IPNS resolution at https://codepen.io/SgtPooki/pen/ZEPrJYp?editors=1011

```
Log:
dnslink:query: "app.aave.com"
dnslink:query: "app.aave.com"
dnslink:answer: {"Status":0,"TC":false,"RD":true,"RA":true,"AD":true,"CD":false,"Question":[{"name":"app.aave.com.","type":16}],"Authority":[{"name":"aave.com.","type":6,"TTL":1800,"data":"art.ns.cloudflare.com. dns.cloudflare.com. 2332124099 10000 2400 604800 1800"}],"Comment":"Response from 173.245.59.102."}
dnslink:query: "_dnslink.app.aave.com"
dnslink:answer: {"Status":0,"TC":false,"RD":true,"RA":true,"AD":true,"CD":false,"Question":[{"name":"_dnslink.app.aave.com.","type":16}],"Answer":[{"name":"_dnslink.app.aave.com.","type":16,"TTL":300,"data":"dnslink=/ipfs/bafybeifcaqowoyito3qvsmbwbiugsu4umlxn4ehu223hvtubbfvwyuxjoe/"}],"Comment":"Response from 172.64.32.152."}
dnslink:answer: {"Status":0,"TC":false,"RD":true,"RA":true,"AD":true,"CD":false,"Question":[{"name":"app.aave.com","type":16}],"Authority":[{"name":"aave.com","type":6,"TTL":1800,"data":"art.ns.cloudflare.com. dns.cloudflare.com. 2332124099 10000 2400 604800 1800"}]}
dnslink:cache: "/ipfs/bafybeifcaqowoyito3qvsmbwbiugsu4umlxn4ehu223hvtubbfvwyuxjoe/"

Result: Invalid value
```

packages/verified-fetch/src/verified-fetch.ts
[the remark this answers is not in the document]
SgtPooki: With the current structure of the code, they cannot override DNS/ENS resolution, as the class `VerifiedFetch` is not publicly consumable, and the `createVerifiedFetch` function only supports gateways/routers parameters. We could add later if someone desires, but for now they would need to resolve the CIDs outside of this method, and then query IPFS/IPNS using this library

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+ * @example Using ipfs protocol to fetch an image
+ *
+ * ```typescript
+ * import { createVerifiedFetch } from '@helia/verified-fetch'
+ *
+ * const fetch = await createVerifiedFetch({
+ *  gateways: ['http://mygateway.info', 'http://trustless-gateway.link']
+ * })
+ * const response = await fetch('ipfs://bafyFoo') // CID for some image file
+ * const blob = await response.blob()
```
lidel, line 59: Maybe go extra step?
```suggestion
 * const blob = await response.blob()
 * const img = document.createElement('img')
 * img.src = URL.createObjectURL(blob)
 * document.body.appendChild(img)
```

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+ *
+ * @example Using ipns protocol to fetch a video
+ *
+ * ```typescript
+ * import { createVerifiedFetch } from '@helia/verified-fetch'
+ *
+ * const fetch = await createVerifiedFetch({
+ *  gateways: ['http://mygateway.info', 'http://trustless-gateway.link']
+ * })
+ * const response = await fetch('ipns://mydomain.com/path/to/video.mp4')
```
lidel, line 70: I think we should avoid video here, because videos require efficient seeking, and that can be done with range requests, which we don't have yet.

Let's make this about streaming generic "big file" instead:

```suggestion
 * @example Using ipns protocol to stream a big file
 *
 * ```typescript
 * import { createVerifiedFetch } from '@helia/verified-fetch'
 *
 * const fetch = await createVerifiedFetch({
 *  gateways: ['http://mygateway.info', 'http://trustless-gateway.link']
 * })
 * const response = await fetch('ipns://mydomain.com/path/to/very-long-file.log')
```

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+ * const videoStreamReader = await response.body.getReader()
```
lidel, line 71: nit: I think we forgot to close this code block?
````suggestion
 * const bigFileStreamReader = await response.body.getReader()
``` 
````

File: packages/verified-fetch/src/index.ts [hunk 41]
```diff
@@ -0,0 +1,339 @@
...
+ *
+ *
+ * @example Using a CID instance to fetch JSON
+ *
+ * ```typescript
+ * import { createVerifiedFetch } from '@helia/verified-fetch'
+ * import { CID } from 'multiformats/cid'
+ *
+ * const fetch = await createVerifiedFetch({
+ *  gateways: ['http://mygateway.info', 'http://trustless-gateway.link']
```
lidel, line 42: nit: some examples use `http://`, some use domain alone. I think we should switch everything to `https://` for consistency and to avoid mixed-content warnings.

packages/verified-fetch/src/index.ts
```diff
@@ -0,0 +1,339 @@
...
+ * - IPFS URL
+ * - IPNS URL
+ *
+ * @example
+ *
+ * ```ts
+ * import { createVerifiedFetch } from '@helia/verified-fetch'
+ *
+ * const fetch = await createVerifiedFetch({
+ *  gateways: ['mygateway.info', 'trustless-gateway.link']
```
lidel, line 26: nit: change to `mygateway.example.net` (https://en.wikipedia.org/wiki/Example.com)

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -0,0 +1,269 @@
...
+    const format = this.getFormat({ headerFormat: new Headers(options?.headers).get('accept'), queryFormat: query.format ?? null })
+
+    const formatHandler = this.formatHandlers[format]
+
+    if (formatHandler != null) {
+      response = await formatHandler.call(this, { cid, path, options })
+    }
+
+    if (response == null) {
+      const codecHandler = this.codecHandlers[cid.code]
```
SgtPooki, line 236: Before we call formatHandler or codecHandler, if there is a path, we should be letting `ipfs-unixfs-exporter` handle the walk to the terminal element, and potentially parsing of the terminal element if not handled by `ipfs-unixfs-exporter`

packages/verified-fetch/src/index.ts
SgtPooki: I thought about adding that but ended up getting distracted I guess, good call :)

SgtPooki:
FYI @achingbrain @lidel @aschmahmann that ipfs-unixfs-exporter is now being used for walking paths

achingbrain (approved):

Issue #395, opened 2024-01-18
Valid IPNS records fail to resolve


2color:
Using the canary tag of the `@helia/http`, trying to resolve the following IPNS record ``, results in the following error:

```
Uncaught (in promise) CodeError: Could not find record for routing key
    at DefaultIPNS.findIpnsRecord_fn (@helia_ipns.js?v=3f44548b:28394:11)
    at async DefaultIPNS.resolve (@helia_ipns.js?v=3f44548b:28319:20)
    at async resolveIpns (index.ts:36:10)
    at async HTMLButtonElement.<anonymous> (index.ts:53:17)
findIpnsRecord_fn @ @helia_ipns.js?v=3f44548b:28394
Show 1 more frame
Show less
```

[Reproduction](https://codesandbox.io/p/github/2color/helia-browser-examples/main?file=%2Fsrc%2Findex.ts%3A14%2C22-14%2C84&workspaceId=96c32827-0145-4a4e-80e4-9201a3af1db9):

achingbrain:
This has been fixed for me after #396

PR #396, opened 2024-01-18
fix: update ipns module to v9 and fix double verification of records

### LINKED ISSUE #394: IPNS records are validated twice when resolved
## Background

Currently, the @helia/ipns package relies on the routers abstraction to resolve IPNS record. This allows it to use more than one router to resolve a record, e.g. DHT, PubSub, and delegated routing API.

As it currently stands, the IPNS module handles verification of every record it gets from a router: https://github.com/ipfs/helia/blob/19bf9ce9bbd5122f1ee5b97dc200ef1658a40d4f/packages/ipns/src/index.ts#L425-L434

Additionally, the[helia-delegated-routing-v1-http-api](https://github.com/ipfs/helia-delegated-routing-v1-http-api/blob/main/packages/client/src/client.ts#L241) also validates the IPNS record which results in two validations which is unnecessary. 

Would be good to add an option to the https://github.com/ipfs/helia-delegated-routing-v1-http-api to avoid verifying the record

achingbrain:
Updates IPNS module to version that is capable of round-tripping records created with non-nano-second validity times - adds an interop test to ensure there are no regressions.

Updates `@helia/delegated-routing-v1-http-api-client` to version that can skip verification of incoming records.

Fixes #394 and #395


2color (approved):
LGTM

PR #398, opened 2024-01-19
feat: add block session support to @helia/interface

achingbrain:
There are no implementations yet but the usage pattern will be something like:

```javascript
// unixfs cat command
export async function * cat (cid: CID, blockstore: Blocks, options: Partial<CatOptions> = {}): AsyncIterable<Uint8Array> {
  // create a session for the CID if support is available
  const blocks = await (blockstore.createSession != null ? blockstore.createSession(cid, options) : blockstore)
  const opts: CatOptions = mergeOptions(defaultOptions, options)

  // resolve and export using the session, if created, otherwise fall back to regular blockstore access
  const resolved = await resolve(cid, opts.path, blocks, opts)
  const result = await exporter(resolved.cid, blocks, opts)

  if (result.type !== 'file' && result.type !== 'raw') {
    throw new NotAFileError()
  }

  if (result.content == null) {
    throw new NoContentError()
  }

  yield * result.content(opts)
}
```

Alternatively the user can control session creation:

```javascript
import { unixfs } from '@helia/unixfs'
import { createHelia } from 'helia' // or http

const node = await createHelia()
const rootCid = CID.parse('Qmfoo')
const sessionBlockstore = await node.blockstore.createSession(rootCid, {
  signal: AbortSignal.timeout(5000)
})

// all operations will use the same session
const fs = unixfs({ blockstore: sessionBlockstore })

for await (const entry of fs.ls(rootCid) {
  if (entry.type !== 'file') {
    continue
  }

  for await (const buf of fs.cat(entry.cid)) {
    // ...
  }
}
```

Removes the `BlockAnnouncer`/`BlockRetriever` single-method interface `BlockBroker` split because we would have to add another `BlockSessionFactory` interface for this which starts getting unwieldy.  Instead just have all the methods be optional and filter the brokers before use.


packages/interface/src/blocks.ts
```diff
@@ -54,7 +56,18 @@ ProgressOptions<PutBlockProgressEvents>, ProgressOptions<PutManyBlocksProgressEv
...
+   * subset of network peers which respond as having the block for the initial
+   * root CID.
+   *
+   * Any blocks written to the blockstore as part of the session will propagate
+   * to the blockstore the session was created from.
+   *
+   * This method is optional to maintain compatibility with existing
+   * blockstores that do not support sessions.
+   */
+  createSession?(root: CID, options?: AbortOptions & ProgressOptions<GetBlockProgressEvents>): Promise<Blockstore>
```
SgtPooki, line 70: To be clear, the signal passed here is for the entire retrival session?

E.g. Your example would abort if the entire session lasted longer than 5000ms, correct?


```
  const sessionBlockstore = await node.blockstore.createSession(rootCid, {
  signal: AbortSignal.timeout(5000)
})
```

packages/utils/src/utils/networked-storage.ts
```diff
@@ -4,23 +4,19 @@ import filter from 'it-filter'
...
+import type { BlockBroker, Blocks, Pair, DeleteManyBlocksProgressEvents, DeleteBlockProgressEvents, GetBlockProgressEvents, GetManyBlocksProgressEvents, PutManyBlocksProgressEvents, PutBlockProgressEvents, GetAllBlocksProgressEvents, GetOfflineOptions, BlockRetrievalOptions } from '@helia/interface/blocks'
 import type { AbortOptions, ComponentLogger, Logger, LoggerOptions, Startable } from '@libp2p/interface'
 import type { Blockstore } from 'interface-blockstore'
 import type { AwaitIterable } from 'interface-store'
 import type { CID } from 'multiformats/cid'
 import type { MultihashHasher } from 'multiformats/hashes/interface'
 
-export interface GetOptions extends AbortOptions {
-  progress?(evt: Event): void
+export interface NetworkedStorageStorageInit {
```
SgtPooki, line 14:
```suggestion
export interface NetworkedStorageInit {
```
?

packages/utils/src/utils/networked-storage.ts
```diff
@@ -200,6 +196,25 @@ export class NetworkedStorage implements Blocks, Startable {
...
     yield * this.child.getAll(options)
   }
+
+  async createSession (root: CID, options?: AbortOptions & ProgressOptions<GetBlockProgressEvents>): Promise<Blocks> {
+    const blockBrokers = await Promise.all(this.blockBrokers.map(async broker => {
+      if (broker.createSession == null) {
+        return broker
+      }
+
+      return broker.createSession(root, options)
```
SgtPooki, line 206: Not really related to this section of code but: how do these session blockstores get providers from delegated-routing to the sessions?

SgtPooki (approved):
+1 to combining blockBroker method types.

Could you explain how a session would become aware of providers obtained from a delegated-routing call?

packages/interface/src/blocks.ts
```diff
@@ -54,7 +56,18 @@ ProgressOptions<PutBlockProgressEvents>, ProgressOptions<PutManyBlocksProgressEv
 GetOfflineOptions & ProgressOptions<GetBlockProgressEvents>, GetOfflineOptions & ProgressOptions<GetManyBlocksProgressEvents>, ProgressOptions<GetAllBlocksProgressEvents>,
 ProgressOptions<DeleteBlockProgressEvents>, ProgressOptions<DeleteManyBlocksProgressEvents>
 > {
-
+  /**
+   * A session blockstore is a special blockstore that only pulls content from a
+   * subset of network peers which respond as having the block for the initial
+   * root CID.
```
aschmahmann, line 62: Out of curiosity what's the reason for putting the session information in the blockstore rather than passing it along with say the signals?

My concern here (which might not be valid) is that it feels like we did something similar in the Go code and needing users to remember to create new block fetching stacks for every session ended up leading to the sessions not being plumbed through the stack well and therefore not having one "code session" per "logical session".

We're currently trying to alter how we do this to allow it to happen via signal (or "context" in Go). See https://github.com/ipfs/boxo/issues/567 and https://github.com/ipfs/kubo/issues/7198.

Every language/environment is different so maybe these concerns would be less likely to surface here.

packages/interface/src/blocks.ts
SgtPooki: Also, do we want to return `Blockstore` explitly and not `Blocks`?

achingbrain: > To be clear, the signal passed here is for the entire retrival session?

No, the signal is just for the session creation, e.g. when we do a `findProvs` and send `WANTHAVE` to create the initial set of providers.

> Also, do we want to return Blockstore explitly and not Blocks?

It returns `Blockstore` intentionally so that you can't create a session from a session.

achingbrain:
> Could you explain how a session would become aware of providers obtained from a delegated-routing call?

HTTP Delegated Routers are configured as a [router](https://helia.io/interfaces/_helia_utils.HeliaInit.html#routers), and when you use the [routing methods](https://helia.io/interfaces/_helia_interface.index.Helia.html#routing) it'll query them.

E.g.:

```TypeScript
import { createHeliaHttp from } '@helia/http'
import { createDelegatedRoutingV1HttpApiClient } from '@helia/delegated-routing-v1-http-api-client'

const node = createHeliaHttp({
  routers: [
    createDelegatedRoutingV1HttpApiClient('https://example.com')
  ]
})

// this will call the http router internally:
for await (const prov of node.routing.findProviders(cid)) {
  if (prov.protocols.includes('transport-ipfs-gateway-http')) {
    // do http things
  }

  if (prov.protocols.includes('transport-bitswap')) {
    // do bitswap things
  }

  // etc
}
```

packages/utils/src/utils/networked-storage.ts
achingbrain: Similar to https://github.com/ipfs/helia/pull/398#issuecomment-1917049330 - you accept an instance of the Helia [Routing](https://helia.io/interfaces/_helia_interface.index.Routing.html) as a component and query it:

E.g.:

```TypeScript
import { createHeliaHttp from } '@helia/http'
import { createDelegatedRoutingV1HttpApiClient } from '@helia/delegated-routing-v1-http-api-client'
import { Routing } from '@helia/interface'

interface MyBlockBrokerComponents {
  routing: Routing
}

interface MyBlockBrokerInit {
  // broker-specific config
}

class MyBlockBroker {
  private routing: Routing

  constructor (components: MyBlockBrokerComponents, init: MyBlockBrokerInit) {
    this.routing = components.routing
  }
  
  createSession (root: CID): Promise<BlockBroker> {
    for await (const prov of this.routing.findProviders(root)) {
      if (prov.protocols.includes('my-block-broker-protocol')) {
        // do block broker stuff
      }
    }

   // ...more code here
  }
}

function createMyBlockBroker(init: MyBlockBrokerInit) {
  return (components) => new MyBlockBroker(components, init)
}

const node = createHeliaHttp({
  blockBrokers: [
    createMyBlockBroker({
      // ...init stuff here
    })
  ]
  routers: [
    createDelegatedRoutingV1HttpApiClient('https://example.com')
  ]
})

// this will call the broker's createSession method
const sessionBlockstore = await node.blockstore.createSesssion(cid)
```

packages/interface/src/blocks.ts
achingbrain: My feeling is that if people are using the API incorrectly it's probably because the correct usage isn't obvious enough, either through basic things like argument names and positions or documentation or both.

We already have people not passing `AbortSignal`s into long-lived operations so I think a session identifier there would also get missed.

The primary interaction point for people with DAGs of various shapes are the `@helia/*` modules.  Since these are (mostly?) stateless they are essentially disposable so I think pushing the session/no session decision to the user (e.g. do I use `helia.blockstore` or `helia.blockstore.createSession(rootCid)`) and having them interact with separate `@helia/unixfs` instances for each session will create a smaller amount of cognitive load.

packages/utils/src/utils/networked-storage.ts
achingbrain: Fixed

packages/interface/src/blocks.ts
```diff
@@ -54,10 +56,21 @@ ProgressOptions<PutBlockProgressEvents>, ProgressOptions<PutManyBlocksProgressEv
...
+   *
+   * Any blocks written to the blockstore as part of the session will propagate
+   * to the blockstore the session was created from.
+   *
+   * This method is optional to maintain compatibility with existing
+   * blockstores that do not support sessions.
```
SgtPooki, line 68:
```suggestion
```

it's no longer optional

packages/interface/src/blocks.ts
```diff
@@ -67,18 +80,65 @@ export type BlockRetrievalOptions<GetProgressOptions extends ProgressOptions = P
...
+  maxProviders?: number
+
+  /**
+   * When searching for providers of the root CID, implementations can check
+   * that providers are still online and have the requested block. This setting
+   * controls how many peers to query at the same time.
+   *
+   * @default 5
+   */
+  queryConcurrency?: number
```
SgtPooki, line 113:
```suggestion
  providerQueryConcurrency?: number
```

packages/interface/src/blocks.ts
```diff
@@ -67,18 +80,65 @@ export type BlockRetrievalOptions<GetProgressOptions extends ProgressOptions = P
...
+  /**
+   * Create a new session
+   */
+  createSession?(root: CID, options?: CreateSessionOptions<RetrieveProgressEvents>): Promise<BlockBroker<RetrieveProgressEvents, AnnounceProgressEvents>>
 }
 
-export type BlockBroker = BlockRetriever | BlockAnnouncer
+export const DEFAULT_SESSION_MIN_PROVIDERS = 1
+export const DEFAULT_SESSION_MAX_PROVIDERS = 5
+export const DEFAULT_SESSION_QUERY_CONCURRENCY = 5
```
SgtPooki, line 143:
```suggestion
export const DEFAULT_SESSION_PROVIDER_QUERY_CONCURRENCY = 5
```

packages/utils/src/utils/networked-storage.ts
```diff
@@ -222,38 +237,49 @@ export const getCidBlockVerifierFunction = (cid: CID, hasher: MultihashHasher):
...
 
+  const retrievers: Array<Required<Pick<BlockBroker, 'retrieve'>>> = []
+
+  for (const broker of blockBrokers) {
+    if (broker.retrieve != null) {
+      // @ts-expect-error retrieve may be undefined even though we've just
+      // checked that it isn't
+      retrievers.push(broker)
+    }
+  }
```
SgtPooki, line 254: an `isRetrievingBlockBroker` type guard function would be safer:

```suggestion
  for (const broker of blockBrokers) {
    if (isRetrievingBlockBroker(broker)) {
      retrievers.push(broker)
    }
  }
```

packages/utils/src/utils/networked-storage.ts
```diff
@@ -200,6 +196,25 @@ export class NetworkedStorage implements Blocks, Startable {
...
+    return new NetworkedStorage({
+      blockstore: this.child,
+      blockBrokers,
+      hashers: this.hashers,
+      logger: this.logger
+    }, {
+      root
+    })
+  }
 }
```
SgtPooki, line 219:
```suggestion

function isRetrievingBlockBroker (broker: BlockBroker): broker is Required<Pick<BlockBroker, 'retrieve'>> {
  if (broker.retrieve != null) {
    return true
  }
  return false
}
```

SgtPooki (approved):
approved, but left some comments for improvement

packages/utils/src/utils/networked-storage.ts
```diff
@@ -88,9 +84,9 @@ export class NetworkedStorage implements Blocks, Startable {
 
     options.onProgress?.(new CustomProgressEvent<CID>('blocks:put:providers:notify', cid))
 
-    this.blockAnnouncers.forEach(provider => {
-      provider.announce(cid, block, options)
-    })
+    await Promise.all(
+      this.components.blockBrokers.map(async broker => broker.announce?.(cid, block, options))
+    )
```
SgtPooki, line 89: do we want allSettled instead of all? do we want a `put` to fail if a single announcement fails?

SgtPooki (approved):

SgtPooki:
@achingbrain anything blocking the merging of this? were you able to test this successfully in helia-http-gateway or should I take a look at that?

packages/utils/src/utils/networked-storage.ts
achingbrain: Perhaps `allSettled`.  Maybe we'd want to throw if all blocker provers failed to announce?

PR #401, opened 2024-01-23
fix!: remove gossipsub from default libp2p services

achingbrain:
It's not required to IPFS things, it's a large chunk of code and it's quite CPU intensive so do not bundle it by default.

People can still configure themselves if they need it.

### DIFF
--- a/packages/helia/package.json
+++ b/packages/helia/package.json
@@ -54,7 +54,6 @@
     "prepublishOnly": "node scripts/update-version.js && npm run build"
   },
   "dependencies": {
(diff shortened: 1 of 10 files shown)


2color:
Nice. Do we track the bundle size at all?

achingbrain:
Not actively, we used to rely on https://bundlephobia.com for this but it [doesn't work with exports maps](https://github.com/pastelsky/bundlephobia/issues/639) 😭

achingbrain:
Just discovered https://packagephobia.com/ - this one seems to work:

- https://packagephobia.com/result?p=helia
- https://packagephobia.com/result?p=@helia/http

PR #412, opened 2024-02-02
fix: @helia/verified-fetch init args are optional

achingbrain:
To allow no-option, all-defaults construction of verified fetch

```TypeScript
import { createVerifiedFetch } from '@helia/verified-fetch'

const fetch = await createVerifiedFetch()

const resp = await fetch('ipfs://bafy...')
// ...
```

### DIFF
--- a/packages/verified-fetch/src/index.ts
+++ b/packages/verified-fetch/src/index.ts
@@ -278,15 +278,15 @@ export interface VerifiedFetchInit extends RequestInit, ProgressOptions<BubbledP
 /**
  * Create and return a Helia node
  */
-export async function createVerifiedFetch (init: Helia | CreateVerifiedFetchWithOptions): Promise<VerifiedFetch> {
+export async function createVerifiedFetch (init?: Helia | CreateVerifiedFetchWithOptions): Promise<VerifiedFetch> {
   if (!isHelia(init)) {
     init = await createHeliaHTTP({
       blockBrokers: [
         trustlessGateway({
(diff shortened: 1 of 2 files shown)


packages/verified-fetch/src/index.ts
```diff
@@ -278,15 +278,15 @@ export interface VerifiedFetchInit extends RequestInit, ProgressOptions<BubbledP
...
  * Create and return a Helia node
  */
-export async function createVerifiedFetch (init: Helia | CreateVerifiedFetchWithOptions): Promise<VerifiedFetch> {
+export async function createVerifiedFetch (init?: Helia | CreateVerifiedFetchWithOptions): Promise<VerifiedFetch> {
   if (!isHelia(init)) {
     init = await createHeliaHTTP({
       blockBrokers: [
         trustlessGateway({
-          gateways: init.gateways
+          gateways: init?.gateways
```
SgtPooki, line 286: just confirming for myself. if gateways is null, should be set to defaults via https://github.com/ipfs/helia/blob/f243de26eda64951c2909121730b6a1b8a689bf6/packages/block-brokers/src/trustless-gateway/broker.ts#L21

SgtPooki (approved):

packages/verified-fetch/src/index.ts
achingbrain: Yeah, the routers don't work in quite the same way though maybe they should.

PR #413, opened 2024-02-02
feat: allow use as a singleton

achingbrain:
This might make things even easier to get started?  It exports a singleton version of verified fetch that creates the backing Helia node on the first fetch access, transparent to the user.

```TypeScript
import { verifiedFetch } from '@helia/verified-fetch'

const response = await verifiedFetch('ipfs://Qmfoo')

console.info(await response.json())
```

It's not in this PR but maybe we'd want to allow some config:

```TypeScript
import { verifiedFetch } from '@helia/verified-fetch'

const response = await verifiedFetch('ipfs://Qmfoo', {
  gateways: ['https://...']
})

console.info(await response.json())
```

Calling this twice with different gateways might result in an error.

```TypeScript
import { verifiedFetch } from '@helia/verified-fetch'

const response1 = await verifiedFetch('ipfs://Qmfoo', {
  gateways: ['https://foo...']
})

const response2 = await verifiedFetch('ipfs://Qmfoo', {
  gateways: ['https://bar...']
})
// Error: Only one set of gateways/routers may be specified, please use `createVerifiedFetch` for more flexibility
```

### DIFF
--- a/packages/verified-fetch/src/singleton.ts
+++ b/packages/verified-fetch/src/singleton.ts
@@ -1,0 +1,23 @@
+import { createVerifiedFetch } from './index.js'
+import type { Resource, VerifiedFetch, VerifiedFetchInit } from './index.js'
+
+interface VerifiedFetchSingleton extends VerifiedFetch {
+  _impl?: VerifiedFetch
+}
+
+const singleton: VerifiedFetchSingleton = async function verifiedFetch (resource: Resource, options?: VerifiedFetchInit): Promise<Response> {
+  if (singleton._impl == null) {
+    singleton._impl = await createVerifiedFetch()
+  }
+
+  return singleton._impl(resource, options)
+}
+singleton.start = async function () {
+  await singleton._impl?.stop()
+}
+singleton.stop = async function () {
+  await singleton._impl?.stop()
+}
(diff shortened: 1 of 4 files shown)


packages/verified-fetch/src/singleton.ts
```diff
@@ -0,0 +1,23 @@
...
+
+const singleton: VerifiedFetchSingleton = async function verifiedFetch (resource: Resource, options?: VerifiedFetchInit): Promise<Response> {
+  if (singleton._impl == null) {
+    singleton._impl = await createVerifiedFetch()
+  }
+
+  return singleton._impl(resource, options)
+}
+singleton.start = async function () {
+  await singleton._impl?.stop()
```
2color, line 16:
```suggestion
  await singleton._impl?.start()
```

SgtPooki (approved):

SgtPooki:
> It's not in this PR but maybe we'd want to allow some config:


I think we should leave the singleton to the most basic, unconfigurable, usecase and have users use createVerifiedFetch if they want to customize gateways, so that we don't have to muck around with error and edge-cases or confusion on the second options argument.

SgtPooki (approved):
love the updates to the readme.

PR #417, opened 2024-02-05
fix: use blockstore interface where possible

achingbrain:
Instead of depending on the Helia `Blocks` interface, depend on `Blockstore` as none of the extra methods are used.

This improves compatibility with other modules and allows reuse from non-Helia contexts.


SgtPooki (approved):
code changes look fine. 

Q: Are we sure we don't want an abstraction layer? I get the type benefits for now, but future extensibility could be hindered by not having an abstraction layer? 

counterpoint: I can't think of how we would want to extend the blockstore other than your sessionBlockstore.

achingbrain:
> future extensibility could be hindered by not having an abstraction layer?

Perhaps, but YAGNI and all that.  We can re-evaluate if requirements change in the future.

PR #426, opened 2024-02-09
fix: content-type response header hints how to process response

achingbrain:
## Description

DAG-JSON and DAG-CBOR can have embeded CIDs which are deserialized to CID objects by the `@ipld/dag-json` and `@ipld/dag-cbor` modules.

This fixes a bug whereby we were JSON.stringifying them which lead to rubbish being returned from `.json()`.

## Notes & open questions

This overwrites the `.json` method on the returned `Response` object.  Is that bad?

## Change checklist


packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -113,7 +113,11 @@ export class VerifiedFetch {
       onProgress: options?.onProgress
     })
     options?.onProgress?.(new CustomProgressEvent<CIDDetail>('verified-fetch:request:end', { cid: cid.toString(), path }))
-    const response = new Response(JSON.stringify(result), { status: 200 })
+    // return body as binary
+    const body = await this.helia.blockstore.get(cid)
+    const response = new Response(body, { status: 200 })
+    // return pre-parsed object with embedded CIDs as objects
+    response.json = async () => result
```
achingbrain, line 120: If this `Response` is [cached](https://developer.mozilla.org/en-US/docs/Web/API/Cache) this might break?

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -97,31 +97,19 @@ describe('@helia/verifed-fetch', () => {
 
   describe('implicit format', () => {
     let verifiedFetch: VerifiedFetch
-    let unixfsStub: ReturnType<typeof stubInterface<UnixFS>>
-    let dagJsonStub: ReturnType<typeof stubInterface<DAGJSON>>
-    let jsonStub: ReturnType<typeof stubInterface<HeliaJSON>>
-    let dagCborStub: ReturnType<typeof stubInterface<DAGCBOR>>
+    let unixfsStub: StubbedInstance<UnixFS>
```
achingbrain, line 100: Much simpler types.

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -97,31 +97,19 @@ describe('@helia/verifed-fetch', () => {
...
+    let dagJsonStub: StubbedInstance<DAGJSON>
+    let jsonStub: StubbedInstance<HeliaJSON>
+    let dagCborStub: StubbedInstance<DAGCBOR>
     let pathWalkerStub: SinonStub<Parameters<PathWalkerFn>, ReturnType<PathWalkerFn>>
-    let blockstoreStub: ReturnType<typeof stubInterface<Blocks>>
+    let blockstoreStub: StubbedInstance<Blocks>
 
     beforeEach(async () => {
       blockstoreStub = stubInterface<Blocks>()
-      unixfsStub = stubInterface<UnixFS>({
```
achingbrain, line 109: `stubInterface` makes every property access return a `sinon.stub()` so this is not necessary.

packages/verified-fetch/src/verified-fetch.ts
achingbrain: I guess we either want this, or we want to convert `CID` objects to the magic `{ "/": "QmFoo" }` structure?

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -310,13 +298,27 @@ describe('@helia/verifed-fetch', () => {
...
+      dagJsonStub.get.withArgs(cid).returns(Promise.resolve({
+        hello: 'world',
+        link: CID.parse('QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN')
+      }))
+      const resp = await verifiedFetch.fetch(cid)
+      const data = await resp.json()
+      expect(data).to.deep.equal({
+        hello: 'world',
+        link: CID.parse('QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN')
+      })
```
achingbrain, line 312: This may be better, the CID encoded as a [link](https://ipld.io/specs/codecs/dag-json/spec/#links):

```suggestion
      expect(data).to.deep.equal({
        hello: 'world',
        link: {
          '/': 'QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN'
        }
      })
```

SgtPooki: I like this, along with https://github.com/ipfs/helia/pull/426#discussion_r1484525834

2color: Really naive question, but wouldn't the CID end up having two different object instances (even if it has the same underlying cid) causing the deep equal test to fail?

packages/verified-fetch/src/verified-fetch.ts
2color: > I think we should alter the body itself, so when cached we return the same thing

What do you mean by alter? 

Either way, if you were to alter, would you first need to parse the dag-json?

packages/verified-fetch/test/verified-fetch.spec.ts
achingbrain: Yeah, there would be two instances but the test would pass because deep equal compares every property of the two objects rather than asserting they are the same instance.

lidel: `dag-json` is returned by gateways as `application/json`-compatible string. It has no concept of non-string representation of CID in JSON, and I don't think we should introduce a custom CID type in `.json()` response.

The way `.json()` works on the web is that it never adds custom types beyond standard JSON. If  `verified-fetch` aims to be a solid foundation to build on the web, we can't break that expectation. 

(+ usual bias: based on years of experience we had with JS ecosystem, I don't think we can confidently say that CID type introduced here won't break in a few months)

Please let's switch to string-only, as suggested by @achingbrain  above:

```js
expect(data).to.deep.equal({
        hello: 'world',
        link: {
          '/': 'QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN'
        }
})
```

If we want typed returns, that would could be introduced in the future as a separate abstraction, `.ipld()` or similar.

lidel (changes requested):
I strongly believe we should not leak custom types onto object returned by `.json()`. See https://github.com/ipfs/helia/pull/426/files#r1485686844

achingbrain:
> I strongly believe we should not leak custom types onto object returned by .json()

I agree, though we now have a weird DX niggle where these return incompatible types:

```js
import { verifiedFetch } from '@helia/verified-fetch'
import { dagJson } from '@helia-dag-json'

const cid = CID.parse('QmDAG-JSONwithCID')

// obj has CID as `{ "/": "QmFoo" }`
const res = await verifiedFetch(`ipfs://${cid}`)
const obj = await res.json()

// obj has CID as object
const d = dagJson(helia)
const obj = await d.get(CID.parse('QmDAG-JSONwithCID')
```

`DAG-CBOR` is more problematic when trying to treat is as plain-old-JSON because it supports several unserializable types:

```js
const obj = {
  hello: 'world',
  bigInt: 12345n,
  bytes: Uint8Array.from([0, 1, 2, 3, 4])
}

const d = dagCbor(helia)
const cid = await d.add(obj)

const res = await verifiedFetch(`ipfs://${cid}`)
const output = await res.json() // 💥
```

I guess the thing to do here is to convert `CID`s to `{ "/": "Qmfoo"}`,  `BigInt`s to strings, `Uint8Array` to plain arrays, and so on - then `.json()` would work.

We could add a `.object()` method to the response type that returns rich object types?

It wouldn't survive [caching](https://developer.mozilla.org/en-US/docs/Web/API/Cache), but we can't cache `ipfs://QmFoo` URLs without browser support anyway.

SgtPooki:
> It wouldn't survive [caching](https://developer.mozilla.org/en-US/docs/Web/API/Cache), but we can't cache ipfs://QmFoo URLs without browser support anyway.

With verified-fetch in a SW (helia-service-worker-gateway), the returned responses are cached because its the Response object itself that's cached by the service-worker, based on requested URL. I'm not 100% sure of the intricacies there, but the returned object will be whatever is in the body i'm sure, so we need to make sure the stream/arrayBuffer given to the `Response` constructor is what we're depending on being cached.

This also brings up a few targeted use cases: 

1. applications that expect the browser to cache the response object of verified-fetch (SW & other gateway-esque solutions)
2. applications that expect an easy fetch interface and subsequent consumption of the returned data.

> It'd be nice to have arrayBuffer() return the raw CBOR, then the user could pass it to @ipld/dag-cbor to get a rich object back, but we can't have that and have .json() return an object as for that to work the response body has to be set as a JSON string so arrayBuffer() would return a array buffer containing that JSON string and not CBOR - at that point all the CBOR codepoints that delimit types like Uint8Array etc will have been lost so there's no way to convert it back to the original object.

I think having `.arrayBuffer()` for dagCbor be the priority makes the most sense, given that the data in the response should be unique per requested URL, and consistent. If someone calls `.json` on dagCbor, it should fail, right? Afterall, it's not actual JSON ™. If we need a codec that is actual JSON, not some psuedo-json-like-object, we should have a different codec, shouldn't we?

Maybe we could provide some interface for verifiedFetch similar to `hashers` for Helia, where we can extend how we process returned content? I imagine handling all of the possible scenarios could easily become painful.

lidel:
I think getting DAG-CBOR experience with `.json()` and `.arrayBuffer()` is extra important, because it is a common pattern for people. We've been telling people to build apps that use a functional equivalent of  `ipfs dag put`. By default, the command turns JSON into CBOR: accepts json string, parses it as dag-json, and stores it as dag-cbor:

```console
$ echo -n '{"hello":"world"}' | ipfs dag put # [--store-codec=dag-cbor] [--input-codec=dag-json]
bafyreidykglsfhoixmivffc5uwhcgshx4j465xwqntbmu43nb2dzqwfvae
```

So if someone builds an app, and wants to read that dag-cbor back with verified-fetch, they would want `.json()` to return JSON similar to what was passed to `ipfs dag put`:

```js
> resp.json()
{"hello":"world"}
```

For practical purposes the roundtrip works fine, as long it is limited to safe subset of CBOR and JSON, listed in 
- https://ipld.io/specs/codecs/dag-cbor/spec/ and
- https://ipld.io/specs/codecs/dag-json/spec/ where tldr is:
  - String keys only, sorted, no whitespace.
  - Avoid floats/Numbers, use strings.
  - Binary arrays are represented as base64 strings: `{"/": { "bytes": String /* Base64 encoded binary */ }}`

I think if we have to choose between `.arrayBuffer()` and `.json()` for a DAG-CBOR CID, having JSON result will be more useful to devs.

@achingbrain Is my assumption correct that if someone wants to get raw `dag-cbor` bytes to parse  with `@ipld/dag-cbor`, we could implement verified-fetch so they can make an explicit request with `Accept` header set to `application/vnd.ipld.dag-cbor` (or `application/vnd.ipld.raw`), and then `.arrayBuffer()` would work fine and return CBOR bytes, ano not JSON string, right?

achingbrain:
Yes, they can request the raw block and decode it themselves, then they can use the rich types and we don't have to worry about breaking `.json()`.

----

Notes from the dApps working group call:

We can try to deserialize the CBOR as JSON and error if it fails.

The user can then re-request with the header `Accept: application/vnd.ipld.raw` to get raw bytes and use `@ipld/dag-cbor` to decode the `.arrayBuffer()` output.

Presumably if they request with `Accept: application/vnd.ipld.dag-cbor, application/vnd.ipld.raw` (e.g. both types) we'll try to decode and if it fails, fall back to letting them use `.arrayBuffer()`.

achingbrain:
That said, what if we just try to decode and just let the content type of the response dictate how it can be consumed, to save the user making two requests?

If the block decodes as JSON-compliant DAG-CBOR it's `application/json`, otherwise it's `application/octet-stream`.

```js
import { verifiedFetch } from '@helia/verified-fetch'
import * as dagCbor from '@ipld/dag-cbor'

const response = await verifiedFetch('ipfs://bafyDagCbor')

let obj

if (response.headers.get('Content-Type')?.includes('application/json')) {
  obj = await response.json()
} else {
  obj = dagCbor.decode(await response.arrayBuffer())
}
```

achingbrain:
Ok, I think we're in a good place now.

## JSON codec

```TypeScript
import { verifiedFetch } from '@helia/verified-fetch'
import * as json from 'multiformats/codecs/json'

const res = await verifiedFetch('ipfs://bafyJSONCodec')

// get object representation of JSON body
console.info(await res.json()) // { ... }

// alternative way to do the same thing
const obj = json.decode(new Uint8Array(await res.arrayBuffer()))
console.info(obj) // { ... }
```

## DAG-JSON codec

```TypeScript
import { verifiedFetch } from '@helia/verified-fetch'
import * as dagJson from '@ipld/dag-json'

const res = await verifiedFetch('ipfs://bafyDAGJSONCodec')

// get spec-compliant plain-old-JSON object that happens to contain a CID
console.info(await res.json()) // { cid: { '/': 'Qmfoo' } }

// ...or use @ipld/dag-json to get rich-object version
const obj = dagJson.decode(new Uint8Array(await res.arrayBuffer()))
console.info(obj) // { cid: CID('Qmfoo') }
```

## DAG-CBOR codec

```TypeScript
import { verifiedFetch } from '@helia/verified-fetch'
import * as dagCbor from '@ipld/dag-cbor'

const res = await verifiedFetch('ipfs://bafyDAGCBORCodec')

if (res.headers.get('content-type') === 'application/json') {
  // CBOR was JSON-friendly
  console.info(await res.json()) // { ... }
} else {
  // CBOR was not JSON-friendly, must use @ipld/dag-cbor to decode
  const obj = dagCbor.decode(new Uint8Array(await res.arrayBuffer()))
  console.info(obj) // { ... }
}
```

If the `DAG-CBOR` block contains anything that cannot round-trip to JSON (e.g. `CID`s, `Uint8Array`s, `BigInt`s, etc), the content type will be `application/octet-stream` - `.json()` will throw and so `@ipld/dag-cbor` must be used to decode the resolved value of `.arrayBuffer()`.

----

I've added README docs explaining the above, please comment if they muddy the waters.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -120,42 +113,35 @@ export class VerifiedFetch {
...
-    const result = await this.dagCbor.get<Uint8Array>(cid, {
-      signal: options?.signal,
-      onProgress: options?.onProgress
-    })
+    // return body as binary
+    const block = await this.helia.blockstore.get(cid)
+    let body: string | Uint8Array
+
+    try {
+      body = dagCborToSafeJSON(block)
```
2color, line 137: This looks like an expensive call

achingbrain: Also the price we pay for trying to turn a "better JSON" back into JSON.. 😅

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -120,42 +109,36 @@ export class VerifiedFetch {
...
-    const response = okResponse(JSON.stringify(result))
-    response.headers.set('content-type', 'application/json')
-    return response
-  }
-
   private async handleJson ({ cid, path, options }: FetchHandlerFunctionArg): Promise<Response> {
     this.log.trace('fetching %c/%s', cid, path)
     options?.onProgress?.(new CustomProgressEvent<CIDDetail>('verified-fetch:request:start', { cid, path }))
-    const result: Record<any, any> = await this.json.get(cid, {
+    const result = await this.helia.blockstore.get(cid, {
```
achingbrain, line 115: We can skip the decode step here because the CID has already told us this is either JSON or DAG-JSON so we can let the `Response` object do the parsing when/if `.json()` is called.

2color: Where is the decode actually happening here? Does the blockstore just store the bytes unparsed?

achingbrain: There is nothing to decode as such - a `Uint8Array` block for the `JSON` and `DAG-JSON` codecs just contains a unicode JSON string, much like the response body.  So much so that we just set the block as the response body then everything is parsable and cacheable via the native `Response` object without any extra handling.

If we want a rich object (e.g. with `CID` and `Uint8Array` objects) we'd use the `@ipld/dag-json` codec with the output of `.arrayBuffer`.

Does that make sense?

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -210,6 +239,34 @@ describe('@helia/verifed-fetch', () => {
...
+      const obj = {
+        hello: 'world'
+      }
+      const j = json(helia)
+      const cid = await j.add(obj)
+
+      const resp = await verifiedFetch.fetch(cid)
+      expect(resp.headers.get('content-type')).to.equal('application/json')
+
+      const output = ipldJson.decode(new Uint8Array(await resp.arrayBuffer()))
```
achingbrain, line 266: After https://github.com/multiformats/js-multiformats/pull/287 we can drop the `Uint8Array` wrapper:

```suggestion
      const output = ipldJson.decode(await resp.arrayBuffer())
```

packages/verified-fetch/src/index.ts
```diff
@@ -138,6 +138,188 @@
...
+ * const obj = dagJson.decode(new Uint8Array(await res.arrayBuffer()))
+ * console.info(obj.cid) // CID(baeaaac3imvwgy3zao5xxe3de)
+ * console.info(obj.buf) // Uint8Array(5) [ 0, 1, 2, 3, 4 ]
+ * ```
+ *
+ * #### DAG-CBOR
+ *
+ * [DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) uses the [Concise Binary Object Representation](https://cbor.io/) format for serialization instead of JSON.
+ *
+ * This supports more datatypes in a safer way than JSON and is smaller on the wire to boot so is usually preferable to JSON or DAG-JSON.
```
SgtPooki, line 291: 🎉

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -81,14 +76,11 @@ export class VerifiedFetch {
...
   private readonly unixfs: HeliaUnixFs
-  private readonly dagJson: DAGJSON
-  private readonly dagCbor: DAGCBOR
-  private readonly json: JSON
   private readonly pathWalker: PathWalkerFn
   private readonly log: Logger
   private readonly contentTypeParser: ContentTypeParser | undefined
 
-  constructor ({ helia, ipns, unixfs, dagJson, json, dagCbor, pathWalker }: VerifiedFetchComponents, init?: VerifiedFetchInit) {
+  constructor ({ helia, ipns, unixfs, pathWalker }: VerifiedFetchComponents, init?: VerifiedFetchInit) {
```
SgtPooki, line 83: this is great. The more we can stay out of pulling in all of the extra codecs, the better.

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -235,9 +370,134 @@ describe('@helia/verifed-fetch', () => {
...
+      }
+      const c = dagCbor(helia)
+      const cid = await c.add(obj)
+
+      const resp = await verifiedFetch.fetch(cid)
+      expect(resp.headers.get('content-type')).to.equal('application/json')
+
+      const output = ipldDagCbor.decode(new Uint8Array(await resp.arrayBuffer()))
+      await expect(c.add(output)).to.eventually.deep.equal(cid)
+    })
```
SgtPooki, line 483: do we need to call this out in packageDocumentation (or is it already?)

Is this forever not-possible, or something we hope to support in the future? (we could remove this test, or assert that it isn't equal)

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -235,9 +370,134 @@ describe('@helia/verifed-fetch', () => {
...
+
+      const resp = await verifiedFetch.fetch(cid)
+      expect(resp.headers.get('content-type')).to.equal('application/json')
+
+      const output = await resp.json()
+      await expect(c.add(output)).to.eventually.deep.equal(cid)
+    })
+
+    // N.b. this is not possible because the incoming block is turned into JSON
+    // and returned as the response body, so `.arrayBufer()` returns a string
```
SgtPooki, line 469:
```suggestion
    // and returned as the response body, so `.arrayBuffer()` returns a string
```

SgtPooki (approved):
Approved, but I imagine we'll wait on https://github.com/multiformats/js-multiformats/pull/287. Looks amazing, fantastic work as always.

packages/verified-fetch/README.md
```diff
@@ -140,14 +140,198 @@ import { fileTypeFromBuffer } from '@sgtpooki/file-type'
...
+
+This supports more datatypes in a safer way than JSON and is smaller on the wire to boot so is usually preferable to JSON or DAG-JSON.
+
+##### Content-Type
+
+Not all data types supported by `DAG-CBOR` can be successfully turned into JSON and back.
+
+When a decoded block can be round-tripped to JSON, the `Content-Type` will be set to `application/json`. In this case the `.json()` method on the `Response` object can be used to obtain an object representation of the response.
+
+When it cannot, the `Content-Type` will be `application/octet-stream` - in this case the `@ipld/dag-json` module must be used to deserialize the return value from `.arrayBuffer()`.
```
SgtPooki, line 311: It might be worth calling out that `.arrayBuffer()` cannot simply be used to then parse as JSON? i.e. https://github.com/ipfs/helia/pull/426#discussion_r1489929304

packages/verified-fetch/src/verified-fetch.ts
2color: Nice! Yes makes total sense.

achingbrain:
> In its current state, there's no way to opt out of the automatic JSON parsing

If they know it's going to fail they can add an `Accept: application/vnd.ipld.raw` to the request and it'll just be a raw block with the content type `application/octet-stream` which they can `await res.arrayBuffer()` into `@ipld/dag-cbor`?

2color:
> If they know it's going to fail they can add an Accept: application/vnd.ipld.raw to the request and it'll just be a raw block with the content type application/octet-stream which they can await res.arrayBuffer() into @ipld/dag-cbor?

From a look at the code, this isn't implemented and the example you provided returns a response object with status 501.

https://github.com/ipfs/helia/blob/f2c0f35ad32f6705a19b82479d4dbc72c9465858/packages/verified-fetch/test/verified-fetch.spec.ts#L98-L113

Another thing which might be a bug is that we pass the options object to a method which expects the body as the first parameter:


https://github.com/ipfs/helia/blob/f2c0f35ad32f6705a19b82479d4dbc72c9465858/packages/verified-fetch/src/verified-fetch.ts#L68-L73

https://github.com/ipfs/helia/blob/f2c0f35ad32f6705a19b82479d4dbc72c9465858/packages/verified-fetch/src/verified-fetch.ts#L296


I also tried the example and it doesn't work

packages/verified-fetch/README.md
```diff
@@ -140,14 +140,198 @@ import { fileTypeFromBuffer } from '@sgtpooki/file-type'
...
+
+[DAG-CBOR](https://ipld.io/docs/codecs/known/dag-cbor/) uses the [Concise Binary Object Representation](https://cbor.io/) format for serialization instead of JSON.
+
+This supports more datatypes in a safer way than JSON and is smaller on the wire to boot so is usually preferable to JSON or DAG-JSON.
+
+##### Content-Type
+
+Not all data types supported by `DAG-CBOR` can be successfully turned into JSON and back.
+
+When a decoded block can be round-tripped to JSON, the `Content-Type` will be set to `application/json`. In this case the `.json()` method on the `Response` object can be used to obtain an object representation of the response.
```
achingbrain, line 309:
```suggestion
When a decoded block can be round-tripped to JSON, the response body will be a JSON string and the `Content-Type` header will be set to `application/json`. In this case the `.json()` method on the `Response` object can be used to obtain an object representation of the response and the `.arrayBuffer()` method will return the JSON string as a byte array - note that this will not be parsable by the `@ipld/dag-cbor` module.
```

packages/verified-fetch/README.md
achingbrain, line 311:
```suggestion
When it cannot, the `Content-Type` will be `application/octet-stream` - in this case the `@ipld/dag-json` module must be used to deserialize the return value from `.arrayBuffer()` and the `.json()` method cannot be used since the response body will be unparsed `CBOR` bytes.
```

packages/verified-fetch/test/verified-fetch.spec.ts
achingbrain: I left this in because it's the logical next test give the input/output matrix, but it won't work for reasons outlined in the comment.  We can remove it if we think it's redundant.

packages/verified-fetch/test/verified-fetch.spec.ts
achingbrain: This has been done now

packages/verified-fetch/README.md
achingbrain: Done!

achingbrain:
I'm going to merge this so we can proceed with the tests for the different data types in the examples.

PR #434, opened 2024-02-14
fix: verified-fetch etag header

SgtPooki:
Fixes some issues with the etag header

~~FYI: `should handle identity CID` test is failing for me locally, even on `main` branch. no idea why. I `npm run clean && npm i && npm run test` and still get the failure~~ I needed to `npm run clean && npm i` from the monorepo root.


lidel (approved):
Thanks, this Etag should be good enough and not cause trouble.

packages/verified-fetch/src/utils/parse-url-string.ts
```diff
@@ -1,6 +1,7 @@
 import { peerIdFromString } from '@libp2p/peer-id'
 import { CID } from 'multiformats/cid'
 import { TLRU } from './tlru.js'
+import type { RequestFormatShorthand } from '../types'
```
achingbrain, line 4:
```suggestion
import type { RequestFormatShorthand } from '../types.js'
```

SgtPooki: thanks. missed this one

SgtPooki (commented):
self review. should have done this yesterday. my bad. should be good to go now

achingbrain (approved):

PR #438, opened 2024-02-16
feat: support `Accept` header in `@helia/verified-fetch`

achingbrain:
Let users get raw data back from CIDs that would otherwise trigger decoding as JSON or CBOR etc by specifying an `Accept` header.

```typescript
const res = await verifiedFetch(cid, {
  headers: {
    accept: 'application/octet-stream'
  }
})
console.info(res.headers.get('accept')) // application/octet-stream
```

Make sure the content-type matches the accept header:

```typescript
const res = await verifiedFetch(cid, {
  headers: {
    accept: 'application/vnd.ipld.raw'
  }
})
console.info(res.headers.get('accept')) // application/vnd.ipld.raw
```

Support multiple values, match the first one:

```typescript
const res = await verifiedFetch(cid, {
  headers: {
    accept: 'application/vnd.ipld.raw, application/octet-stream, */*'
  }
})
console.info(res.headers.get('accept')) // application/vnd.ipld.raw
```

If they specify an Accept header we can't honour, one that doesn't match the eventual content type, or one returned by the content type parser return a [406](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406):

```typescript
const res = await verifiedFetch(cid, {
  headers: {
    accept: 'application/what-even-is-this'
  }
})
console.info(res.status) // 406
```

## Change checklist


packages/verified-fetch/test/accept-header.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+  it('should suuport wildcards', async () => {
+    const obj = {
+      hello: 'world'
+    }
+    const c = dagCbor(helia)
+    const cid = await c.add(obj)
+
+    const resp = await verifiedFetch.fetch(cid, {
+      headers: {
+        accept: 'application/what-even-is-this, */*'
```
lidel, line 90: nit: we were not testing with `;q=`, below adds what Firefox sends, should be enough as a regression test:

```suggestion
        accept: 'application/what-even-is-this,text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8'
```

lidel (approved):

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -523,4 +512,89 @@ describe('@helia/verifed-fetch', () => {
...
+        headers: {
+          accept: 'application/octet-stream'
+        }
+      })
+      expect(resp.headers.get('content-type')).to.equal('application/octet-stream')
+      const output = ipldDagCbor.decode(new Uint8Array(await resp.arrayBuffer()))
+      expect(output).to.deep.equal(obj)
+    })
+
+    it('should return a 406 if the content cannot be represented by the mime type in the accept header', async () => {
```
2color, line 552: I thought that this would be allowed as `CID` links can be seamlessly converted to JSON and back without loss.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -134,6 +181,12 @@ export class VerifiedFetch {
     try {
       body = dagCborToSafeJSON(block)
```
2color, line 182: How would this behave for `dag-cbor` block with JSON compliant data **and** a CID link?

packages/verified-fetch/test/verified-fetch.spec.ts
achingbrain: The existing `DAG-CBOR` + `CID` test asserts that it fails to parse and the content-type header is `application/octet-stream` - https://github.com/ipfs/helia/blob/main/packages/verified-fetch/test/verified-fetch.spec.ts#L377-L390

That could be incorrect behaviour, which we can change, but what do you think of the UX here?

It feels a bit weird I think because if it parsed as `JSON` you'd have a `DAG-CBOR` block with a `CID` in it which we effectively translate to `DAG-JSON` (so the body is `application/json`).  If you wanted to get `CID` objects in the response you'd have to use `@ipld/dag-json` to decode it, even though you used a `DAG-CBOR` `CID`.

packages/verified-fetch/src/verified-fetch.ts
achingbrain: See https://github.com/ipfs/helia/pull/438#discussion_r1494771739

Right now a `CID` in a `DAG-CBOR` block is not JSON compliant.

packages/verified-fetch/test/verified-fetch.spec.ts
achingbrain: As an example:

```js
// an object with a CID
const objWithCid = {
  link: CID.parse('Qmfoo')
}

// store as DAG-CBOR
const dagCborCID = await heliaDagCbor.put(objWithCid)
const block = await helia.blockstore.get(dagCborCID)
console.info(block) // 101010101

// store as DAG-JSON
const dagJsonCID = await heliaDagJson.put(objWithCid)
const block = await helia.blockstore.get(dagJsonCID)
console.info(block) // { "link": { "/": "Qmfoo" } }

// fetch using DAG-CBOR CID
const res = await verifiedFetch(dagCborCID)

// if we succeed parsing it as JSON
console.info(res.headers.get('content-type')) // application-json

// can deserialize using `.json()` on `Response` object
console.info(await res.json()) // { "link": { "/": "Qmfoo" } }

// cannot decode as DAG-CBOR any more using `.arrayBuffer()`
const obj = ipldDagCbor.decode(await res.arrayBuffer()) // explodes - because the block is effectively now a JSON string

// we have to use DAG-JSON now
const obj = ipldDagJson.decode(await res.arrayBuffer())
```

How obvious do you think the final step is?  Do we care that it's

2color: > It feels a bit weird to me because if it parsed as JSON you'd have a DAG-CBOR block with a CID in it which we effectively translate to DAG-JSON (so the body is application/json). If you wanted to get CID objects in the response you'd have to use @ipld/dag-json to decode it, even though you used a DAG-CBOR CID.

I agree with you. Thanks for pointing that out.

--- 

Also your example shows why it doesn't make sense:
```
// This isn't obvious at all... 
const obj = ipldDagJson.decode(await res.arrayBuffer())
```

2color (approved):
Looks good. 

Nice to have [the challenge I brought up](https://github.com/ipfs/helia/pull/426#issuecomment-1946049671) addressed by this. 💯

achingbrain:
@lidel @2color please take another look - I've added a lot more tests, support for wildcards and q-factor and also the IPLD serialized type conversions from https://specs.ipfs.tech/http-gateways/path-gateway/#accept-request-header

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -68,21 +74,63 @@ function okResponse (body?: BodyInit | null): Response {
...
+  'application/vnd.ipld.raw',
+  'application/octet-stream'
+]
+
+/**
+ * if the user has specified an `Accept` header, and it's in our list of
+ * allowable "raw" format headers, use that instead of detecting the content
+ * type, to avoid the user signalling that they will Accepting one mime type
+ * and then receiving something different.
+ */
```
SgtPooki, line 106:
```suggestion
/**
 * if the user has specified an `Accept` header, and it's in our list of
 * allowable "raw" format headers, use that instead of detecting the content
 * type. This avoids the user from receiving something different when they 
 * signal that they want to `Accept` a specific mime type.
 */
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -92,60 +140,123 @@ export class VerifiedFetch {
...
+      body = block
+    }
+
+    const response = okResponse(body)
+    response.headers.set('content-type', accept ?? 'application/json')
     return response
   }
 
-  private async handleDagCbor ({ cid, path, options }: FetchHandlerFunctionArg): Promise<Response> {
+  private async handleDagCbor ({ cid, path, accept, options }: FetchHandlerFunctionArg): Promise<Response> {
```
SgtPooki, line 201: we might need to break these functions out at some point as well.

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -92,60 +140,123 @@ export class VerifiedFetch {
...
+    let terminalElement: UnixFSEntry | undefined
+    let ipfsRoots: CID[] | undefined
+
+    try {
+      const pathDetails = await walkPath(this.helia.blockstore, `${cid.toString()}/${path}`, options)
+      ipfsRoots = pathDetails.ipfsRoots
+      terminalElement = pathDetails.terminalElement
+    } catch (err) {
+      this.log.error('Error walking path %s', path, err)
+      // return new Response(`Error walking path: ${(err as Error).message}`, { status: 500 })
```
SgtPooki, line 257:
```suggestion
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -185,19 +294,27 @@ export class VerifiedFetch {
     const response = okResponse(stream)
     await this.setContentType(firstChunk, path, response)
 
-    options?.onProgress?.(new CustomProgressEvent<CIDDetail>('verified-fetch:request:end', { cid: resolvedCID, path: '' }))
+    if (ipfsRoots != null) {
+      response.headers.set('X-Ipfs-Roots', ipfsRoots.map(cid => cid.toV1().toString()).join(',')) // https://specs.ipfs.tech/http-gateways/path-gateway/#x-ipfs-roots-response-header
+    }
```
SgtPooki, line 299: we will likely need to pull this out of `handleDagPb` so we can set roots for others as well

SgtPooki (approved):
a few comments, but we could fix those later if necessary. this is a huge net win.

SgtPooki (approved):
a few comments, but we could fix those later if necessary. this is a huge net win.

packages/verified-fetch/src/utils/select-output-type.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const CID_TYPE_MAP: Record<number, string[]> = {
+  [dagCborCode]: [
+    'application/json',
+    'application/vnd.ipld.dag-cbor',
+    'application/cbor',
+    'application/vnd.ipld.dag-json',
+    'application/octet-stream',
+    'application/vnd.ipld.raw',
+    'application/vnd.ipfs.ipns-record',
```
2color, line 20: A bit late to the party, but how can `'application/vnd.ipfs.ipns-record',` be a valid response type for the `dag-cbor`, `dag-json` and the others that have it included?

packages/verified-fetch/test/accept-header.spec.ts
```diff
@@ -78,7 +156,23 @@ describe('accept header', () => {
...
+    const c = dagCbor(helia)
+    const cid = await c.add(obj)
+
+    const resp = await verifiedFetch.fetch(cid, {
+      headers: {
+        accept: 'application/what-even-is-this, */*, application/vnd.ipld.raw'
+      }
+    })
+    expect(resp.status).to.equal(200)
+    expect(resp.headers.get('content-type')).to.equal('application/json')
```
2color, line 172: This is pretty counter intuitive, but my understanding is that the order matters and that when processing `*/*` the default mime-type is returned, which according to the map is `application/json`

packages/verified-fetch/src/utils/select-output-type.ts
SgtPooki: good catch. it seems like ipns-records are only supposed to be dag-cbor? https://specs.ipfs.tech/ipns/ipns-record/, but the way this is implemented, someone can request an ipns-record and get back json, dag-json, etc.. 

I'm not sure if someone would want to, but it seems like this would allow users to get a non-ipld-ish/different-ipld-ish data object from an ipns-record?

achingbrain: It’s valid for the CID type. So I request and IPNS name, that resolves to a DAG-CBOR CID, but I send an Accept header for the IPNS record. Not straightforward, it can probably be simplified.

2color: Got it! Yeah that makes sense.

PR #441, opened 2024-02-20
feat: support downloading car files from @helia/verified-fetch 

achingbrain:
Adds support for the `application/vnd.ipld.car` accept header to allow downloading CAR files of DAGs.

## Change checklist


packages/verified-fetch/test/car.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+    })
+    expect(resp.status).to.equal(200)
+    expect(resp.headers.get('content-type')).to.equal('application/vnd.ipld.car; version=1')
+    expect(resp.headers.get('content-disposition')).to.equal(`attachment; filename="${cid.toString()}.car"`)
+    const buf = new Uint8Array(await resp.arrayBuffer())
+
+    expect(buf).to.equalBytes(await writer.bytes())
+  })
+
+  it('should support specify a filename for a CAR file', async () => {
```
SgtPooki, line 49:
```suggestion
  it('should support specifying a filename for a CAR file', async () => {
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -157,7 +160,27 @@ export class VerifiedFetch {
...
+      .catch(err => {
+        this.log.error('could not write car', err)
+        stream.cancel(err)
+          .catch(err => {
+            this.log.error('could not cancel stream after car export error', err)
+          })
+      })
+
+    const response = okResponse(stream)
+    response.headers.set('content-type', 'application/vnd.ipld.car; version=1')
```
SgtPooki, line 181: Do we need to check/set the version of the car file we get/create?

https://specs.ipfs.tech/http-gateways/trustless-gateway/#car-version

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -157,7 +160,27 @@ export class VerifiedFetch {
...
+
+    // write the DAG behind `cid` into the writer
+    c.export(cid, writer, options)
+      .catch(err => {
+        this.log.error('could not write car', err)
+        stream.cancel(err)
+          .catch(err => {
+            this.log.error('could not cancel stream after car export error', err)
+          })
+      })
```
SgtPooki, line 178: I'm not sure I could explain deeply what's going on here.

Could you add a jsdoc link to explainers for me and potential future devs?

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -157,7 +160,27 @@ export class VerifiedFetch {
...
-    return notSupportedResponse('vnd.ipld.car support is not implemented')
+    const c = car(this.helia)
+    const { writer, out } = CarWriter.create(cid)
+
+    const stream = toBrowserReadableStream<Uint8Array>(async function * () {
+      yield * out
+    }())
+
+    // write the DAG behind `cid` into the writer
+    c.export(cid, writer, options)
```
SgtPooki, line 171: We may not want to pass full options here?

SgtPooki (approved):
Lgtm but a few comments and questions

packages/verified-fetch/src/verified-fetch.ts
achingbrain: From what I can tell `@ipld/car`'s writer doesn't add the v2 pragma or append the index so it's only ever going to be a v1 car.  It also doesn't export a method that lets us obtain the version programatically.

I've opened this to double check: https://github.com/ipld/js-car/issues/161

packages/verified-fetch/src/verified-fetch.ts
2color: Same 🙊.

achingbrain: Take a look now - I've simplified the code here and pushed the complexity into the `@helia/car` module which is arguably where it should have been in the first place.

https://github.com/ipfs/helia/pull/444 will need to be merged and this rebased before CI will pass.

Hopefully it's clearer now.

packages/verified-fetch/src/verified-fetch.ts
achingbrain: We can change it if we need to, my experience has generally been that sometimes it's useful to be able to pass options to sub components.

As it stands there's complete overlap between the option types so I don't think all we'd gain by limiting the option keys is verbosity.

packages/verified-fetch/src/verified-fetch.ts
SgtPooki: it's definitely better. Thanks!

SgtPooki (approved):
we have a few pending PRs which may change this PR further, but looks really good.

PR #442, opened 2024-02-21
feat: download tars from @helia/verified-fetch

achingbrain:
Adds support for downloading tar archives of UnixFS directories


packages/verified-fetch/test/tar.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+    const fs = unixfs(helia)
+    const cid = await fs.addBytes(file)
+
+    const resp = await verifiedFetch.fetch(`ipfs://${cid}?format=tar`)
+    expect(resp.status).to.equal(200)
+    expect(resp.headers.get('content-type')).to.equal('application/x-tar')
+    expect(resp.headers.get('content-disposition')).to.equal(`attachment; filename="${cid.toString()}.tar"`)
+  })
+
+  it('should support specify a filename for a TAR file', async () => {
```
SgtPooki, line 124:
```suggestion
  it('should support specifying a filename for a TAR file', async () => {
```

packages/verified-fetch/test/tar.spec.ts
```diff
@@ -0,0 +1,149 @@
...
+    await expect(toBuffer(entries[1].body)).to.eventually.deep.equal(directory[1].content)
+
+    expect(entries[2]).to.have.nested.property('header.name', `${importResult.cid}/${directory[2].path?.split('/')[0]}`)
+
+    expect(entries[3]).to.have.nested.property('header.name', `${importResult.cid}/${directory[2].path}`)
+    await expect(toBuffer(entries[3].body)).to.eventually.deep.equal(directory[2].content)
+
+    expect(entries[4]).to.have.nested.property('header.name', `${importResult.cid}/${directory[0].path}`)
+    await expect(toBuffer(entries[4].body)).to.eventually.deep.equal(directory[0].content)
+  })
```
SgtPooki, line 111: the order here seems odd.. should `${importResult.cid}/${directory[0].path}` be at entries[1] ?

SgtPooki: Alex let me know it-tar handles ordering and it's stable, so it should be fine.

packages/verified-fetch/src/utils/get-tar-stream.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (file.type === 'file' || file.type === 'directory') {
+    mode = file.unixfs.mode
+    mtime = file.unixfs.mtime != null ? new Date(Number(file.unixfs.mtime.secs * 1000n)) : undefined
+  }
+
+  return {
+    name: file.path,
+    mode,
+    mtime,
+    size: Number(file.size),
```
SgtPooki, line 24: do we want to squash the bigint from `Exportable.size` to a Number here?

SgtPooki (approved):
lgtm

packages/verified-fetch/src/utils/get-tar-stream.ts
achingbrain: We have to convert it because `it-tar` expects the field to be a number.

packages/verified-fetch/test/tar.spec.ts
achingbrain: There's some interesting discussion on tar file ordering here: https://unix.stackexchange.com/questions/120143/how-is-the-order-in-which-tar-works-on-files-determined

Either way, it's external to `@helia/verified-fetch`.

PR #444, opened 2024-02-21
feat: stream car file bytes from @helia/car

achingbrain:
To better support streaming CAR files with a less confusing API, add a method to `@helia/car` that takes root CIDs and returns an AsyncGenerator that yields CAR file bytes.

## Change checklist

### DIFF
--- a/packages/car/src/index.ts
+++ b/packages/car/src/index.ts
@@ -61,6 +61,7 @@
 import { CarWriter } from '@ipld/car'
 import drain from 'it-drain'
 import map from 'it-map'
+import merge from 'it-merge'
 import defer from 'p-defer'
 import PQueue from 'p-queue'
 import type { DAGWalker } from '@helia/interface'
@@ -174,7 +175,7 @@ class DefaultCar implements Car {
   }
 
   async export (root: CID | CID[], writer: Pick<CarWriter, 'put' | 'close'>, options?: AbortOptions & ProgressOptions<GetBlockProgressEvents>): Promise<void> {
-    const deferred = defer()
+    const deferred = defer<Error | undefined>()
     const roots = Array.isArray(root) ? root : [root]
 
     // use a queue to walk the DAG instead of recursion so we can traverse very large DAGs
@@ -185,6 +186,7 @@ class DefaultCar implements Car {
       deferred.resolve()
     })
     queue.on('error', (err) => {
+      queue.clear()
       deferred.reject(err)
     })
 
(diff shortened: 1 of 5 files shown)


SgtPooki (approved):
Nice

packages/car/src/index.ts
```diff
@@ -129,6 +130,31 @@ export interface Car {
...
    */
   export(root: CID | CID[], writer: Pick<CarWriter, 'put' | 'close'>, options?: AbortOptions & ProgressOptions<GetBlockProgressEvents>): Promise<void>
+
+  /**
+   * Returns an AsyncGenerator that yields CAR file bytes.
+   *
+   * @example
+   *
+   * ```typescript
+   * import fs from 'node:fs'
```
2color, line 141:
```suggestion
```

SgtPooki: good eye

packages/car/src/index.ts
```diff
@@ -129,6 +130,31 @@ export interface Car {
...
+   * import { Readable } from 'stream'
```
SgtPooki, line 142:
```suggestion
```

packages/car/src/index.ts
```diff
@@ -129,6 +130,31 @@ export interface Car {
...
+   * import { createHelia } from 'helia'
+   * import { car } from '@helia/car
+   * import { CID } from 'multiformats/cid'
+   * import pEvent from 'p-event'
```
SgtPooki, line 146:
```suggestion
```

packages/car/src/index.ts
achingbrain: Done!

packages/car/src/index.ts
achingbrain: Done!

packages/car/src/index.ts
```diff
@@ -170,12 +195,47 @@ class DefaultCar implements Car {
...
+        for await (const buf of out) {
+          yield buf
+        }
+      })(),
+      // eslint-disable-next-line require-yield
+      (async function * () {
+        await deferred.promise
+      })()
+    )
   }
```
SgtPooki, line 239: AFAIK, the `deferred` wont be resolved until `out` is fully done, and `out` finishing will resolve the deferred.promise. Do we need to `await deferred.promise`?

achingbrain: Yes, it's there so if exporting fails the error doesn't get smothered.

packages/car/src/index.ts
```diff
@@ -129,6 +131,31 @@ export interface Car {
...
   export(root: CID | CID[], writer: Pick<CarWriter, 'put' | 'close'>, options?: AbortOptions & ProgressOptions<GetBlockProgressEvents>): Promise<void>
+
+  /**
+   * Returns an AsyncGenerator that yields CAR file bytes.
+   *
+   * @example
+   *
+   * ```typescript
+   * import fs from 'node:fs'
+   * import { Readable } from 'stream'
```
2color:
```suggestion
```

packages/car/src/index.ts
achingbrain: Maybe not, it's to make a test that simulates a write error failing, but thinking about it, if there's an error there it's a bug in `@ipld/car` which would need fixing.  Have removed.

PR #445, opened 2024-02-21
feat: customize ipns dnsResolvers

SgtPooki:
## Title

feat(verified-fetch): customize ipns dnsResolvers

## Description

Adds the ability to customize dns resolvers to verified-fetch.

Users should then be able to add custom doh resolvers to helia-service-worker-gateway as mentioned at https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/22

## Change checklist


SgtPooki (commented):
self review

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -523,4 +523,31 @@ describe('@helia/verifed-fetch', () => {
...
+      const verifiedFetch = new VerifiedFetch({
+        helia
+      }, {
+        dnsResolvers: [customResolver1, customResolver2]
+      })
+      try {
+        await verifiedFetch.fetch('ipns://mydomain.com', { onProgress })
+      } catch {
+        // ignoring error of walking the CID because we haven't actually added the block to the blockstore
+      }
```
achingbrain, line 545: If we expect this to reject, it's worth ensuring it does reject and in the expected way:

```suggestion
      // ignoring error of walking the CID because we haven't actually added the block to the blockstore
      await expect(verifiedFetch.fetch('ipns://mydomain.com', { onProgress })).to.eventually.be.rejected
        .with.property('code', 'ERR_EXPECTED_ERR_CODE')
```

packages/verified-fetch/src/verified-fetch.ts
```diff
@@ -32,6 +32,7 @@ interface VerifiedFetchComponents {
  */
 interface VerifiedFetchInit {
   contentTypeParser?: ContentTypeParser
+  dnsResolvers?: DNSResolver[]
```
achingbrain, line 35: We've used strings for config in other places, do we want to do that here too and do the `string` -> `DNSResolver` conversion for the user?

SgtPooki: I had it as strings first and then swapped to the object.. but a string does make usage easier on devs. Will update.

SgtPooki: oh. the reason i used the object: there's no way to know if its Json or Https resolver..

packages/verified-fetch/test/index.spec.ts
```diff
@@ -51,4 +52,23 @@ describe('createVerifiedFetch', () => {
...
   })
+
+  it('can be passed a contentTypeParser', async () => {
+    const contentTypeParser = (_bytes: Uint8Array): string => 'application/json'
+    const verifiedFetch = await createVerifiedFetch(undefined, {
+      contentTypeParser
+    })
+    expect(verifiedFetch).to.be.ok()
+    await verifiedFetch.stop()
+  })
```
SgtPooki, line 63: this was missing, but i don't think it's adding much

packages/verified-fetch/test/index.spec.ts
```diff
@@ -51,4 +52,23 @@ describe('createVerifiedFetch', () => {
...
+
+  it('can be passed custom dnsResolvers', async () => {
+    const dnsResolver = dnsOverHttps('https://example.com/dns-query')
+    const dnsJsonResolver = dnsJsonOverHttps('https://example.com/dns-json')
```
SgtPooki, line 67: confirm that both types can be passed successfully

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -523,4 +523,23 @@ describe('@helia/verifed-fetch', () => {
       await expect(resp.text()).to.eventually.equal('hello world')
     })
   })
+
+  describe('custom dns-resolvers', () => {
```
SgtPooki, line 527: do we want this in it's own file similar to `content-type-parser.spec.ts`?

SgtPooki (commented):
self review. ready for peers

packages/verified-fetch/test/verified-fetch.spec.ts
```diff
@@ -523,4 +523,23 @@ describe('@helia/verifed-fetch', () => {
...
+
+      customDnsResolver.returns(Promise.resolve('/ipfs/QmVP2ip92jQuMDezVSzQBWDqWFbp9nyCHNQSiciRauPLDg'))
+
+      const verifiedFetch = new VerifiedFetch({
+        helia
+      }, {
+        dnsResolvers: [customDnsResolver]
+      })
+      // error of walking the CID/dag because we haven't actually added the block to the blockstore
+      await expect(verifiedFetch.fetch('ipns://some-non-cached-domain.com')).to.eventually.be.rejected.with.property('errors').that.has.lengthOf(0)
```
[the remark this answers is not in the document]
achingbrain: FWIW, `AggregateError` has really poor DX as the default message logged to the console doesn't include any of the actual errors which can lead to spurious issues being reported and a lot of wasted time.

In other places where there's only one entry in `.errors` we unwind it to just throw that error instead.

In this case, if there's an error, it should be made available, otherwise no error should be thrown?

packages/verified-fetch/test/index.spec.ts
achingbrain: Yes, the test should at least verify that the content type parse is invoked?

packages/verified-fetch/test/verified-fetch.spec.ts
achingbrain: Probably, the test suite gets quite unwieldy otherwise, then it's not obvious to newcomers where to add more tests.

packages/verified-fetch/test/index.spec.ts
achingbrain: Similar to the `contentTypeParser` test, this should confirm that the resolvers are invoked during an appropriate fetch.

packages/verified-fetch/src/index.ts
```diff
@@ -138,6 +138,29 @@
...
+ *
+ * const fetch = await createVerifiedFetch({
+ *  gateways: ['https://trustless-gateway.link'],
+ *  routers: ['http://delegated-ipfs.dev']
+ * }, {
+ *  dnsResolvers: [
+ *    dnsJsonOverHttps('https://my-dns-resolver.example.com/dns-json'),
+ *    dnsOverHttps('https://my-dns-resolver.example.com/dns-query')
+ *  ]
+ * })
```
achingbrain, line 161: Tiny nit but two-space indentation consistent with the rest of the codebase:

```suggestion
 * const fetch = await createVerifiedFetch({
 *   gateways: ['https://trustless-gateway.link'],
 *   routers: ['http://delegated-ipfs.dev']
 * }, {
 *   dnsResolvers: [
 *     dnsJsonOverHttps('https://my-dns-resolver.example.com/dns-json'),
 *     dnsOverHttps('https://my-dns-resolver.example.com/dns-query')
 *   ]
 * })
```

We have `doc-check` but it checks markdown blocks in `README.md` - you need to run `npm run docs` to regenerate `README.md` first, even then I don't think it lints them.

packages/verified-fetch/test/index.spec.ts
SgtPooki: these tests were just confirming construction and I didn't want to change the pattern, but yea we should.

packages/verified-fetch/src/index.ts
```diff
@@ -138,6 +138,29 @@
...
+ *
+ * const fetch = await createVerifiedFetch({
+ *   gateways: ['https://trustless-gateway.link'],
+ *   routers: ['http://delegated-ipfs.dev']
+ * }, {
+ *   dnsResolvers: [
+ *     dnsJsonOverHttps('https://my-dns-resolver.example.com/dns-json'),
+ *     dnsOverHttps('https://my-dns-resolver.example.com/dns-query')
+ *   ]
+ * })
```
SgtPooki, line 161: Another question here. Do we want to add dnsResolvers to the first object?

to stick to the pattern of "components in first object, options that alter how things work in the second" I think it would make sense in either place.. 🤔

2color: On the other hand, customising dnsResolvers requires importing the constructors for dnsOverHttps and dnsJsonOverHttps which does reduce ergonomics.

SgtPooki: it'll still be optional either way, but I think I agree with you.

packages/verified-fetch/test/index.spec.ts
SgtPooki: FYI: the customDnsResolvers are being tested when constructing the `VerifiedFetch` class directly, do we need to add them here?

packages/verified-fetch/src/index.ts
```diff
@@ -138,6 +138,29 @@
  * })
  * ```
  *
+ * ### Custom DNS resolvers
+ *
+ * If you don't want to leek DNS queries to the default resolvers, you can provide your own list of DNS resolvers to `createVerifiedFetch`.
```
achingbrain, line 143:
```suggestion
 * If you don't want to leak DNS queries to the default resolvers, you can provide your own list of DNS resolvers to `createVerifiedFetch`.
```

packages/verified-fetch/src/index.ts
achingbrain: I think maybe in the first argument, keeping the routers/gateways/resolvers together makes sense to me.

The only thing we have in the second arg is `contentTypeParser`, not 100% on why we have two args right now anyway!

SgtPooki: After looking at this for a bit and starting to change both args to a single object, but contentTypeParser has to be in the second obj, or we need to accept a possible `.helia` option in the first obj, or else users can't pass both a custom Helia & contentTypeParser

2color:
I just realised we have another problem now. We cannot instantiate `@helia/verified-fetch` with a custom instance of Helia and also customising `dnsResolver`.

This came up in the SW Gateway, where we now are faced with a choice:
1.  Either we instantiate Helia with the `IDBBlockstore` in the service worker but have no way to customise the dnsResolvers
2. We customise the dnsResolvers (along with gateways and routers) but cannot use the IDBBlockstore.

This is how they compare:
```ts
// option 1
  const helia = await createHeliaHTTP({
    blockstore,
    datastore,
    blockBrokers: [
      trustlessGateway({
        gateways: [...config.gateways, ]
      })
    ],
    routers: [...config.routers, 'https://delegated-ipfs.dev'].map(rUrl => delegatedHTTPRouting(rUrl))
  })

  const verifiedFetch = await createVerifiedFetch(helia, { contentTypeParser })

// ---- OR

// option 2
  const verifiedFetch = await createVerifiedFetch({
    gateways: [...config.gateways, 'https://trustless-gateway.link'],
    routers: [...config.routers, 'https://delegated-ipfs.dev'],
    dnsResolvers: ['https://delegated-ipfs.dev/dns-query'].map(dnsJsonOverHttps)
  }, {
    contentTypeParser
  })
```

2color:
I just realised we have another problem now. We cannot instantiate `@helia/verified-fetch` with a custom instance of Helia and also customising `dnsResolver`.

This came up in the SW Gateway, where we now are faced with a choice:
1.  Either we instantiate Helia with the `IDBBlockstore` in the service worker but have no way to customise the dnsResolvers
2. We customise the dnsResolvers (along with gateways and routers) but cannot use the IDBBlockstore.

This is how they compare:
```ts
// option 1
  const helia = await createHeliaHTTP({
    blockstore,
    datastore,
    blockBrokers: [
      trustlessGateway({
        gateways: [...config.gateways, ]
      })
    ],
    routers: [...config.routers, 'https://delegated-ipfs.dev'].map(rUrl => delegatedHTTPRouting(rUrl))
  })

  const verifiedFetch = await createVerifiedFetch(helia, { contentTypeParser })

// ---- OR

// option 2
  const verifiedFetch = await createVerifiedFetch({
    gateways: [...config.gateways, 'https://trustless-gateway.link'],
    routers: [...config.routers, 'https://delegated-ipfs.dev'],
    dnsResolvers: ['https://delegated-ipfs.dev/dns-query'].map(dnsJsonOverHttps)
  }, {
    contentTypeParser
  })
```


This seems like something we'd want to address. Naively we could move dnsResolvers to the second parameter and frame it as follows: the first config param is to customise Helia or pass a Helia instance, while the second is to customise verified-fetch specific config. 

Any thoughts @SgtPooki @achingbrain ?

PR #448, opened 2024-02-23
fix: only traverse relevant part of HAMT

aschmahmann:
## Description

Switched the UnixFS resolver code to do resolve only the relevant part of a HAMT rather than potentially enumerating all of it.

Fixes: ipfs-shipyard/helia-service-worker-gateway#18

## Change checklist

### DIFF
--- a/packages/unixfs/src/commands/utils/resolve.ts
+++ b/packages/unixfs/src/commands/utils/resolve.ts
@@ -82,7 +82,7 @@ export async function resolve (cid: CID, path: string | undefined, blockstore: B
   log('resolved %s to %c', path, lastCid)
 
   return {
-    cid,
+    cid:lastCid,
     path,
     segments
   }
--- a/packages/unixfs/src/commands/utils/resolve.ts
+++ b/packages/unixfs/src/commands/utils/resolve.ts
@@ -1,11 +1,9 @@
 import { logger } from '@libp2p/logger'
-import { exporter } from 'ipfs-unixfs-exporter'
-import findShardCid from 'ipfs-unixfs-exporter/dist/src/utils/find-cid-in-shard.js'
-import { DoesNotExistError, InvalidParametersError } from '../../errors.js'
+import { walkPath } from 'ipfs-unixfs-exporter'
+import { InvalidParametersError } from '../../errors.js'
 import { addLink } from './add-link.js'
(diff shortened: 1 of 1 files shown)


packages/unixfs/src/commands/utils/resolve.ts
```diff
@@ -61,11 +69,11 @@ export async function resolve (cid: CID, path: string | undefined, blockstore: B
...
-      for await (const entry of result.content()) {
-        if (entry.name === part) {
-          dirCid = entry.cid
-          break
-        }
+      if (result.unixfs?.type === 'hamt-sharded-directory') {
+        // special case - unixfs v1 hamt shards
+        dirCid = await findShardCid(result.node, part, blockstore)
+      } else {
+        dirCid = findLinkCid(result.node, part)
```
aschmahmann, line 76: This code is straight-up copy-pasted from https://github.com/ipfs/js-ipfs-unixfs/blob/4749d9a7c1eddd86b8fc42c3fa47f88c7b1b75ae/packages/ipfs-unixfs-exporter/src/resolvers/unixfs-v1/index.ts#L59 which makes me wonder why we can't just use the exporter or resolver from ipfs-unixfs-exporter directly and pass a path

packages/unixfs/src/commands/utils/resolve.ts
```diff
@@ -1,12 +1,14 @@
 import { logger } from '@libp2p/logger'
 import { exporter } from 'ipfs-unixfs-exporter'
+import findShardCid from 'ipfs-unixfs-exporter/dist/src/utils/find-cid-in-shard.js'
```
aschmahmann, line 3: This isn't right. We could copy-paste or export the code if needed, but wonder if we could use more of the resolver code from ipfs-unixfs-exporter as-is

achingbrain: FWIW you can't do deep imports of files like this any more, it's prevented by the `exports` map in the module's `package.json`.

achingbrain:
I have opened a PR that uses the `unixfs-exporter` to traverse the DAG here - https://github.com/ipfs/helia/pull/455

It should also solve the issue reported in https://github.com/ipfs-shipyard/helia-service-worker-gateway/issues/18

PR #455, opened 2024-02-27
fix: use unixfs exporter to traverse DAGs

achingbrain:
Reuse the existing exporter `walkPath` method to traverse a DAG as it is clever enough to not load unnecessary blocks which is important when traversing HAMTs.

## Change checklist


packages/unixfs/test/cat.spec.ts
```diff
@@ -92,4 +93,42 @@ describe('cat', () => {
...
+    await drain(blockstore.deleteMany([
+      scriptFile.cid,
+      styleFile.cid,
+      imageFile.cid
+    ]))
+
+    // should be able to cat the index file without loading the other files
+    // in the shard - the blockstore is offline so will throw if requested
+    // blocks are not present
+    const bytes = await toBuffer(fs.cat(dir.cid, {
```
SgtPooki, line 128: I see. Nice 👍

packages/unixfs/src/commands/utils/resolve.ts
```diff
@@ -37,57 +38,17 @@ export async function resolve (cid: CID, path: string | undefined, blockstore: B
...
-      })
-    } else {
-      throw new InvalidParametersError('Could not resolve path')
-    }
+  const p = `/ipfs/${cid}${path == null ? '' : `/${path}`}`
+  const segments = await all(walkPath(p, blockstore, options))
+
+  if (segments.length === 0) {
+    throw new DoesNotExistError('Could not find path in directory')
   }
```
SgtPooki, line 46: This is so much cleaner.

SgtPooki (approved):
Lgtm. I knew it should have been easier than Adin and I were making it out to be

PR #459, opened 2024-03-01
feat: add sessions to trustless gateways

achingbrain:
Implements blockstore sessions for trustless gateways.

- Queries the Helia routing for block providers
- Creates a set of trustless gateways from routing results
- Uses only these gateways to fetch session blocks


packages/block-brokers/src/trustless-gateway/broker.ts
```diff
@@ -57,6 +79,85 @@ export class TrustlessGatewayBlockBroker implements BlockBroker<TrustlessGateway
...
+    this.log('finding transport-ipfs-gateway-http providers for cid %c', root)
+
+    const queue = new PeerQueue({
+      concurrency: options.providerQueryConcurrency ?? DEFAULT_SESSION_PROVIDER_QUERY_CONCURRENCY
+    })
+
+    Promise.resolve().then(async () => {
+      for await (const provider of this.routing.findProviders(root, options)) {
+        if (provider.protocols == null || !provider.protocols.includes('transport-ipfs-gateway-http')) {
+          continue
```
lidel, line 107: :thought_balloon:  I worry that as things are today, filtering by `transport-ipfs-gateway-http` will be harmful for decentralization in the long term. 

AFAIK it is something invented by IPNI at cid.contact for Rhea/Saturn last year to allow  nft.storage to avoid bitswap bills. It only works with IPNI PUTs and was not designed with p2p in mind, Amino DHT peers who in the future will announce  `/https` multiaddrs with trustless gateway as alternative to bitswap will not be discoverable this way. 

@achingbrain It may be more future-proof to look at all results, and keep ones that have `Addrs` with multiaddr that has `/https` or `/tls/http`. 
This will be compatible with https://github.com/libp2p/specs/blob/master/http/transport-component.md and will also filter-out useless gateways  without TLS set up (can't use them in browser due to mixed-content error).

If we look at `Addrs` instead, when IPNI+DHT proxy at https://delegated-ipfs.dev/routing/v1/providers/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi starts returning Kubo nodes as  `Schema: peer` with `/https` multiaddr, it will "just work", and helia will be able to act on HTTP transport with these peers without having to change anything.

achingbrain: This is quite sad-face-making. I hadn't realised that `transport-ipfs-gateway-http` was added to the spec due to a commercial concern for `nft.storage` rather than being something you can rely on.

> when IPNI+DHT proxy starts returning DHT results with /https multiaddr, it will "just work"

As I read the above, the HTTP transport being added to Kubo is what will cause HTTP multiaddrs to appear.  Unless I'm missing something it doesn't mean the peer is running a path or subdomain gateway.

That is, the remote could have a HTTP transport, but also have the gateway config disabled, just the presence of the transport isn't enough to say one way or the other.

It seems like the only way is to make a request and see what you get back?  This could be a recipe for DDOSing if you created nodes that respond to provider queries for certain CIDs?

I think a possible solution here might be for the peer to indicate it's capabilities in it's signed peer record?  The HTTP routers could return the peer record in the `GET /providers` response?

alanshaw: I mean FWIW, if it is the case that it was added exclusively for nft.storage then I wasn't aware of that.

As far as I knew the trustless gateway was already a concept, and we implemented it for Rhea when we were asked. I think there was probably a cost saving for us there due to where we were hosting  our bitswap peer at the time, but that's not the case now, and shouldn't have been a convincing argument then since infra costs and locations can easily be changed.

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -36,8 +37,11 @@ export class TrustlessGateway {
    */
   #successes = 0
 
-  constructor (url: URL | string) {
+  private readonly log: Logger
+
+  constructor (url: URL | string, logger: ComponentLogger) {
     this.url = url instanceof URL ? url : new URL(url)
+    this.log = logger.forComponent(`helia:trustless-gateway-block-broker:${this.url.hostname}`)
```
SgtPooki, line 44: this will be nice

SgtPooki (approved):
this looks good to me, but based on prior comments it needs a few changes before merging

packages/block-brokers/src/trustless-gateway/broker.ts
achingbrain: My point is that if you send me a request, and I send you a response knowing it will trigger you to send a request to a specific third party, that is an attack vector.

If I send you a single response with 10x provider records, and you make 10x requests, that is a vector for an amplification attack because I have got you to do more work than the work it took me to get you to do the work, though hopefully it's not all bad because the provider hosts should have to be distinct.

As long as we recognise this and are ok with it.

SgtPooki: Thats a great catch @achingbrain  

We should absolutely have some way to know the returned providers are valid ipfs node.

I guess as long as we dont modify the delegated routers from a set of known good actors it shouldnt be an issue, but thats centralizing on known delegated providers.

I imagine we would want any peer to be able to provide routing responses in the future though (ambient peer discovery?) where this could become a huge problem. I believe the amplification attack was discussed there as well

aschmahmann: > My point is that if you send me a request, and I send you a response knowing it will trigger you to send a request to a specific third party, that is an attack vector.

True, although to some extent this is true regardless. There is currently no validation anywhere in use (e.g. DHT, IPNI, etc.) that a peer advertising `/ip4/1.2.3.4/tcp/4001/p2p/<peerID>` actually controls that IP address so there's already some potential vector there. This could be done by trying each advertised address, although this assumes the routing system supports all the protocols advertised (e.g. ipv4/ipv6, tcp/quic/webtransport/wss/...) which might not be true (or great for upgradability).

To be fair it is likely more work to setup HTTPS for a bad request than trying to establish just a libp2p connection and on the libp2p side we can be somewhat protected by the peerIDs/public keys (e.g. the public key can sign what's a valid advertisement with its key which will cause the security negotiation to fail otherwise ... although that will result in a wasted dial which is a separate issue in and of itself).

While we're here there's a slightly larger attack surface here for "webseeds-like" behavior (e.g. pointing at `https://some-ubuntu-mirror.com/ubuntu-v.1.2.3.iso` or (.car)) since more bytes can be transferred. This problem gets drastically worse if the incremental verifiability window gets larger.

> If we get a list of providers from some delegated-router and all those providers turn out bad, we could at least mark that delegated-router as a bad actor and limit any future requests to it

Hopefully as explained above this isn't a reasonable thing to do unless there's actually responsibility on the underlying routing system to do these checks (or some delegated routers are allowed to advertise via something like HTTP OPTIONS or .well-known to do extra work you can punish them for not doing correctly). Mostly what you're left with to punish them on is do they give about the same results as other delegated-routers (for the same underlying routing systems) or as querying the underlying systems themselves (likely requires something similar to https://github.com/ipfs/specs/pull/388).

aschmahmann: > filtering by transport-ipfs-gateway-http will be harmful for decentralization in the long term.

So this is true, just like filtering for `bitswap` is also harmful. These protocol advertisements are only supported by IPNI, for Amino (as well as any new protocol advertised by nodes without a new codec landing and all IPNI nodes supporting the code) using identify-like behavior (e.g. libp2p identify or leveraging .well-known) is required.

> Since then, I've been trying to find a way for /routing/v1 to represent Amino DHT peers that provide trustless gateway endpoint, ensuring these can be found (once we have them), not just IPNI ones.

For trustless-gateway over libp2p .well-known can be used and it's fine. For plain trustless-gateway there's currently no way to advertise that in the Amino DHT because the mappings are multihash -> peerID. This is very similar (although the underlying use case wasn't exactly the same) as the problem expressed here https://github.com/libp2p/notes/issues/11 and https://github.com/ipld/ipld/issues/57#issuecomment-486867032 (note: that's from 2019 and my understanding of the situation has improved a bit since then 🙃 which is part of why there's more detail in the message above than there). AFAIK this would require a protocol change to support with Amino.

packages/block-brokers/src/trustless-gateway/broker.ts
```diff
@@ -1,26 +1,48 @@
...
 import type { CID } from 'multiformats/cid'
 
+export interface CreateTrustlessGatewaySessionOptions extends CreateSessionOptions<TrustlessGatewayGetBlockProgressEvents> {
+  /**
+   * Specify the cache control header to send to the remote. 'only-if-cached'
+   * will prevent the gateway from fetching the content if they don't have it.
+   *
+   * @default only-if-cached
+   */
+  cacheControl?: string
```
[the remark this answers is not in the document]
achingbrain: This is to stop them doing what gateways do, e.g. fetch content on your behalf.  Otherwise it defeats the purpose of having a session.

PR #470, opened 2024-03-14
chore: correct order in ipns changelog

2color:
Fixes the order of the generated changelog.

Not sure why it wasn't in the right order 🤷‍♂️

## Change checklist


achingbrain (approved):

achingbrain:
Huh, weird.  It's probably to do with using different release tools over time.

PR #471, opened 2024-03-14
feat: add `record`/`answer` fields to IPNS results

achingbrain:
Adds fields to IPNS results:

`record`: present in the return from `ipns.resolve` - contains the resolved IPNS record.

`answer`: present in the return from `ipns.resolveDNSLink` - contains the resolved DNS Answer.

## Change checklist


2color (approved):
Thanks for the quick turnaround! Looks good.

This is a breaking change, right?

achingbrain:
It's non-breaking because we're just adding some more fields to the return types 👍

SgtPooki (approved):

PR #475, opened 2024-03-26
fix: Add GlobSourceResult to globSource return type in unixfs.

saul-jb:
## Title
fix: Add GlobSourceResult to globSource return type.

## Description

This is a small PR that modifies the return type on the `globSource` utility in `@helia/unixfs` to return the `GlobSourceResult` in addition to the `ImportCandidate` type. This means that the `path` property will be a `string` instead of `string | undefined`.

## Notes & open questions

This PR makes it a bit easier to use `globSource` since it is a bit annoying to have to check if `path` is defined when it always is.

Is there any reason why we had the `GlobSourceResult` type but `globSource` does not return it?

## Change checklist


achingbrain:
Thanks!

> Is there any reason why we had the `GlobSourceResult` type but `globSource` does not return it?

No, just an oversight, I think.

PR #484, opened 2024-04-05
fix: increase default listers on abort signals

achingbrain:
To prevent spurious warnings appearing in the console, increase the number of listeners on created abort signals.

## Change checklist

### DIFF
--- a/packages/bitswap/src/bitswap.ts
+++ b/packages/bitswap/src/bitswap.ts
@@ -82,8 +82,8 @@ export class Bitswap implements BitswapInterface {
 
   async want (cid: CID, options: WantOptions = {}): Promise<Uint8Array> {
     const controller = new AbortController()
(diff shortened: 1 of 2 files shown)


SgtPooki (approved):
I would really like us to get to a state where we can manage signals without needing to increase max listeners. But I would also like to be able to customize the max allowed in nodejs. 11 isn't very much and AFAIK there's no best practice for cascading signals like we deal with. 

There is some really good discussion in https://github.com/nodejs/node/issues/35990 r

achingbrain:
I agree it's a shame this is necessary but it seems unlikely to stop being so given that we have a top-level abort signal being passed into a non-trivial set of operations, all of which need to be torn down if the signal fires it's `"abort"` event.

I don't think node should emit this warning, evidence of a memory leak is memory going up and not coming down, not listeners being added to an event emitter.

PR #490, opened 2024-04-09
fix: cancel in-flight block requests when racing brokers

achingbrain:
If multiple brokers are configured, when one resolves a block request,
cancel the outstanding requests to free up resources.

## Change checklist


SgtPooki (commented):
noice

2color:
Nice! I was under the impression we already had this! This is great!

PR #483, opened 2024-04-04
chore: release main

achingbrain:
:robot: I have created a release *beep* *boop*
---

<details><summary>bitswap: 1.0.0</summary>

### Features

* add @helia/bitswap with sessions ([#409](https://github.com/ipfs/helia/issues/409)) ([e582c63](https://github.com/ipfs/helia/commit/e582c63ca296c789312f5fcf5e3e18f267f74c03))

### Bug Fixes

* improve sessions implementation ([#495](https://github.com/ipfs/helia/issues/495)) ([9ea934e](https://github.com/ipfs/helia/commit/9ea934ed7208e87c28bc65e9090bdedf66ceeffd))
* increase default listers on abort signals ([#484](https://github.com/ipfs/helia/issues/484)) ([7cd012a](https://github.com/ipfs/helia/commit/7cd012aa2ba568845d49d63a71806d20f6ac678f))
* remove wants from wantlist when multiple block retrievers are used ([#491](https://github.com/ipfs/helia/issues/491)) ([b1c761d](https://github.com/ipfs/helia/commit/b1c761db6db7a7aca3044263fdd5e8967204deeb))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @helia/interface bumped from ^4.0.0 to ^4.2.0
    * @helia/utils bumped from ^0.1.0 to ^0.2.0
</details>

<details><summary>block-brokers: 2.1.0</summary>

### Features

* add @helia/bitswap with sessions ([#409](https://github.com/ipfs/helia/issues/409)) ([e582c63](https://github.com/ipfs/helia/commit/e582c63ca296c789312f5fcf5e3e18f267f74c03))
* add block session support to @helia/interface ([#398](https://github.com/ipfs/helia/issues/398)) ([5cf216b](https://github.com/ipfs/helia/commit/5cf216baa6806cd82f8fcddd1f024ef6a506f667))
* add sessions to trustless gateways ([#459](https://github.com/ipfs/helia/issues/459)) ([6ddefb0](https://github.com/ipfs/helia/commit/6ddefb01154b970f5ab7ec7cb7445d9eedbc5474))

### Bug Fixes

* improve sessions implementation ([#495](https://github.com/ipfs/helia/issues/495)) ([9ea934e](https://github.com/ipfs/helia/commit/9ea934ed7208e87c28bc65e9090bdedf66ceeffd))

### Dependencies

* The following workspace dependencies were updated
  * dependencies
[...]

PR #492, opened 2024-04-09
chore: order of releases in changelog

2color:


achingbrain:
We've been through a couple of different release tools which format things differently, and code has been moved around without trimming the release logs, it's a bit of a mess.

packages/helia/CHANGELOG.md
```diff
@@ -187,6 +173,19 @@
...
+### Dependencies
+
+* **dev:** bump aegir from 39.0.13 to 40.0.8 ([#198](https://github.com/ipfs/helia/issues/198)) ([4d75ecf](https://github.com/ipfs/helia/commit/4d75ecffb79e5177da35d3106e42dac7bc63153a))
+* update sibling dependencies ([beb10b5](https://github.com/ipfs/helia/commit/beb10b5590d66d1d5bef9b5e890b888263df2c92))
+
+### Dependencies
+
+* The following workspace dependencies were updated
+  * dependencies
+    * @helia/block-brokers bumped from ^2.0.0 to ^2.0.1
```
SgtPooki, line 188: it looks like there's a trailing `dependencies` below this that's actually a part of `### [2.0.1](https://www.github.com/ipfs/helia/compare/helia-v2.0.0...helia-v2.0.1) (2023-08-16)`

SgtPooki (changes requested):
we should probably try to re-generate the changelog completely using semantic-release.

SgtPooki:
@achingbrain if we remove the changelogs in https://github.com/ipfs/helia/pull/483 can we get release-please to regenerate the full changelog?

achingbrain:
I don't know tbh, but my guess would be no.  I think it only considers changes since the last release tag, otherwise we'd see CI releases failing all over the place due to GitHub applying rate limiting (this happens when you stuff up the release-please config, for example).

achingbrain:
Ok, I've updated the formatting of all changelogs to be consistent with the latest release tool.

I've also re-tagged all the old releases from the `@helia/foo-vNNN` format to `foo-vNNN` and removed the old tags so everything's consistent.

Please delete your local tags in any cloned copy of helia you have and fetch the updated list (this will prevent you re-creating the old tag formats if you accidentally run `git push --tags`, for example):

```console
$ git tag -l | xargs git tag -d
$ git fetch
```

Hopefully going forwards everything autogenerated will be done just so.

achingbrain (approved):

PR #493, opened 2024-04-09
fix: wrap blockstore in identity blockstore

achingbrain:
Wraps blockstore to prevent writing identity blocks

## Change checklist


SgtPooki (approved):
TIL why helia was requesting identity CIDs from trustless-gateways.

PR #495, opened 2024-04-12
fix: improve sessions implementation

achingbrain:
- Sessions are created synchronously
- The root CID of a session is filled on the first CID retrieval for zero-delay session creation
- Providers are found and queried for the root block directly, any that have it are added to the session
- Further peers are added to the session as more CIDs are requested
- Further peers are searched for when no current peers have the block for a requested CID
- Providers that have errored (e.g. protocol selection failure) are excluded from the session
- Bitswap only queries provider peers, not directly connected peers
- HTTP Gatways are loaded from the routing
- When providers are returned without multiaddrs we try to load them without blocking yielding of other providers
- Common session code has been moved into an abstract superclass to remove duplication.

## Change checklist


packages/utils/src/routing.ts
```diff
@@ -37,7 +43,17 @@ export class Routing implements RoutingInterface, Startable {
...
+    // back as an empty array - when this happens we have to do a FIND_PEER
+    // query to get updated addresses, but we shouldn't block on this so use a
+    // separate bounded queue to perform this lookup
+    const queue = new PeerQueue<Provider | null>({
+      concurrency: this.providerLookupConcurrency
+    })
+    queue.addEventListener('error', () => {})
+
     for await (const peer of merge(
+      queue.toGenerator(),
```
SgtPooki, line 56: TIL about libp2p utils `peer-queue`. sweet

packages/block-brokers/src/trustless-gateway/session.ts
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        // take first address?
+        // /ip4/x.x.x.x/tcp/31337/http
+        // /ip4/x.x.x.x/tcp/31337/https
+        // etc
+        const uri = multiaddrToUri(httpAddresses[0])
+
+        this.log('found http-gateway provider %p %s for cid %c', provider.id, uri, cid)
+        this.providers.push(new TrustlessGateway(uri, this.logger))
```
SgtPooki, line 83: should we do some filtering here to handle allowInsecure/allowLocal since we're only taking the first multiaddr?

packages/utils/src/abstract-session.ts
```diff
@@ -0,0 +1,221 @@
...
+      }
+
+      // we found it, all good
+      if (foundBlock) {
+        return
+      }
+
+      // find more session peers and retry
+      Promise.resolve()
+        .then(async () => {
```
SgtPooki, line 120: same as elsewhere, why promise.resolve.then().catch() instead of try/catch?

also, linter should be catching this and required a prefixed `void`
```suggestion
      void Promise.resolve()
        .then(async () => {
```

packages/utils/src/abstract-session.ts
```diff
@@ -0,0 +1,221 @@
...
+            deferred.resolve(block)
+          } catch (err: any) {
+            this.log.error('could not find new providers for %c', cid, err)
+            deferred.reject(err)
+          }
+        })
+        .catch(err => {
+          if (options.signal?.aborted === true) {
+            return
+          }
```
SgtPooki, line 138:
```suggestion
          if (options.signal?.aborted === true) {
            // not throwing the aborted signal because xxxxx
            return
          }
```

packages/utils/src/abstract-session.ts
```diff
@@ -0,0 +1,221 @@
...
+
+export interface AbstractCreateSessionOptions extends CreateSessionOptions {
+  name: string
+}
+
+export interface BlockstoreSessionEvents<Provider> {
+  provider: CustomEvent<Provider>
+}
+
+export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents extends ProgressEvent> extends TypedEventEmitter<BlockstoreSessionEvents<Provider>> implements BlockBroker<RetrieveBlockProgressEvents> {
```
SgtPooki, line 24: since this is used by both bitswap and trustlessgateways (and potentially other blockBrokers in the future) we should probably normalize logging messages to use `retrieve`/`announce` instead of `want`/`wanting`

Also, should we add `announce` here at the abstract session level?

packages/utils/src/abstract-session.ts
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * This method should search for providers and emit a `"provider"` event when
+   * they are found.
+   *
+   * The returned promise should resolve when `count` providers have been found
+   * but it should continue to search for providers until the session either has
+   * `this.maxProviders` providers, or the passed abort signal fires an
+   * `"abort"` event.
+   */
+  abstract findNewProviders (cid: CID, count: number, options: AbortOptions): Promise<void>
```
SgtPooki, line 204: we could implement the sentinel value logic in `AbstractSession` so concrete implementations would only have to provide the findProviders logic. Handling the control flow in the "abstract" class would make sure we're not having to maintain the "but it should continue until x or y" logic in multiple places.

1. concrete.findNewProviders -> emits provider events with providers
1. abstract.onProvider -> queries providers unless maxProviders, in which case it emits `stopFindingProviders` 
1. abstract.onAbortedSignal -> emit `stopFindingProviders`
1. concrete.onStopFindingProviders -> stops it's provider routing logic, closes out requests, removes listeners, etc.

This should make it easier to test and reduce duplicated sentinel value handling logic (check for maxProviders, check for abort signal, and anything else we need to do in the future).

packages/utils/src/index.ts
```diff
@@ -101,6 +103,24 @@ export interface HeliaInit {
...
+   * When this happens the peer's info has to be looked up with a further query.
+   *
+   * To not have this query block the yielding of other providers returned with
+   * multiaddrs, a separate queue is used to perform this lookup.
+   *
+   * This config value controls the concurrency of that queue.
+   *
+   * @default 5
+   */
+  providerLookupConcurrency?: number
```
SgtPooki, line 122: beautiful explanation.. very clear

packages/utils/src/utils/networked-storage.ts
```diff
@@ -197,17 +169,54 @@ export class NetworkedStorage implements Blocks, Startable {
     options.onProgress?.(new CustomProgressEvent('blocks:get-all:blockstore:get-many'))
     yield * this.child.getAll(options)
   }
+}
+
+export type NetworkedStorageComponents = StorageComponents
+
+/**
+ * Networked storage wraps a regular blockstore - when getting blocks if the
+ * blocks are not present Bitswap will be used to fetch them from network peers.
```
SgtPooki, line 178:
```suggestion
 * blocks are not present, the configured BlockBrokers will be used to fetch them.
```

SgtPooki (approved):
whew. I'm a big fan of all the listed improvements. Various comments for things I didn't understand and some minor improvements we could make, but overall lgtm

packages/bitswap/src/network.ts
```diff
@@ -263,29 +262,12 @@ export class Network extends TypedEventEmitter<NetworkEvents> {
...
-        }
-      }
+      // make sure we can dial the provider
+      const dialable = await this.libp2p.isDialable(provider.multiaddrs, {
+        runOnTransientConnection: this.runOnTransientConnections
+      })
 
-      // ignore non-bitswap providers
-      if (provider.protocols?.includes('transport-bitswap') === false) {
+      if (!dialable) {
```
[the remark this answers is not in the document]
SgtPooki: we do. found in later code. this is a big PR

packages/bitswap/src/session.ts
```diff
@@ -1,155 +1,141 @@
...
+            detail: prov
           })
+
+          found++
+
+          if (found === count) {
+            this.log('session is ready')
+            deferred.resolve()
+            // continue finding peers until we reach this.maxProviders
+          }
```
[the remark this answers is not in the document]
SgtPooki: it is handled. there is a unit test for this

packages/utils/test/abstract-session.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      session.providers.push({
+        id: await createEd25519PeerId()
+      })
+    })
+    session.queryProvider.withArgs(cid).resolves(block)
+
+    await expect(session.retrieve(cid)).to.eventually.deep.equal(block)
+  })
+
+  it('should request blocks from new providers as they are discovered', async () => {
```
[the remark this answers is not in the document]
SgtPooki: found it and resolved it.

packages/block-brokers/src/bitswap.ts
```diff
@@ -72,16 +72,16 @@ class BitswapBlockBroker implements BlockBroker<BitswapWantBlockProgressEvents,
...
   }
 
-  async createSession (root: CID, options?: CreateSessionOptions<BitswapWantBlockProgressEvents>): Promise<BlockBroker<BitswapWantBlockProgressEvents, BitswapNotifyProgressEvents>> {
-    const session = await this.bitswap.createSession(root, options)
+  createSession (options?: CreateSessionOptions<BitswapWantBlockProgressEvents>): BlockBroker<BitswapWantBlockProgressEvents, BitswapNotifyProgressEvents> {
+    const session = this.bitswap.createSession(options)
 
     return {
       announce: async (cid, block, options) => {
         await this.bitswap.notify(cid, block, options)
```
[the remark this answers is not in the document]
achingbrain: > we want all bitswap peers to know about blocks we have now right?

Yes, if any peers want the block we are announcing (e.g. it's in the want list they've previously sent us), the protocol says we are to send it to them.

> we want any peers we sent a WANT, to receive a HAVE/DONTWANT from us

If we've sent them a `WANT-BLOCK` or `WANT-HAVE`, and we acquire the block by other means before they reply with `HAVE` or `DONT-HAVE`, we send them an updated wantlist cancelling that particular `WANT-*`.

> if we didn't send a WANT/WANTHAVE a want to a peer, and they didn't send a WANT/WANTHAVE to us, we don't necessarily need to notify that peer that we have the block, right?

We'll only notify them if they've previously sent us a `WANT-BLOCK` or a `WANT-HAVE` and they've not subsequently cancelled that `WANT-*`.

packages/block-brokers/src/trustless-gateway/session.ts
```diff
@@ -0,0 +1,157 @@
...
+  includeProvider (provider: TrustlessGateway): boolean {
+    return true
+  }
+
+  sortProviders (providers: TrustlessGateway[]): TrustlessGateway[] {
+    return providers.sort((a, b) => b.reliability() - a.reliability())
+  }
+}
+
+function filterMultiaddrs (multiaddrs: Multiaddr[], allowInsecure: boolean, allowLocal: boolean): Multiaddr[] {
```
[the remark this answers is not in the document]
achingbrain: 🤷 I usually think if the final arg is an object, then it's an options object, and all properties should be optional.

packages/bitswap/src/session.ts
```diff
@@ -1,155 +1,141 @@
...
+      result = await this.wantList.wantSessionBlock(cid, provider.peerId, options)
+    } catch (err: any) {
+      if (err.code === 'ERR_UNSUPPORTED_PROTOCOL') {
+        this.log('%p does not speak bitswap, excluding from session queries', provider.peerId)
+        provider.failed = true
 
-    // TODO findNewProviders when promise.any throws aggregate error and signal
-    // is not aborted
+        // increase max providers so we can find another more suitable peer
+        this.maxProviders++
```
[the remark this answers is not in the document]
achingbrain: The reason this increases is to allow adding extra session peers after the existing peers prove bad or unreliable.

When we add new peers we query the routing for new providers - if this is deterministic we'd get the same set of providers back, so we need a method to exclude known bad peers, which is why this sets a "failed" flag on the peer and allows the session size to increase.

This probably isn't very scalable since we end up storing an unbounded list of bad peers.  I've refactored the code to use a bloom filter to exclude bad peers, this will only ever occupy the memory used by the filter, which is fixed based on the number of hashes it is expected to contain.

packages/bitswap/src/session.ts
```diff
@@ -1,155 +1,141 @@
...
 
-    this.log('find %d-%d new provider(s) for %c', count, this.maxProviders, cid)
-
-    const source = merge(
-      [...this.wantList.peers.keys()],
-      map(this.network.findProviders(cid, options), prov => prov.id)
-    )
-
     void Promise.resolve()
       .then(async () => {
```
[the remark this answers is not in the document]
achingbrain: We want to find `this.minProviders` to allow block retrieval to commence ASAP, but we want the query to continue until we reach `this.maxProviders` to allow pulling from multiple peers at once.

If we used try/catch we'd block on reaching `this.maxProviders` which would slow things down.

packages/bitswap/src/want-list.ts
```diff
@@ -483,7 +465,7 @@ export class WantList extends TypedEventEmitter<WantListEvents> implements Start
         full: true,
         entries: pipe(
           this.wants.entries(),
-          (source) => filter(source, ([key, entry]) => !entry.cancel && (entry.session.size > 0 && !entry.session.has(peerId))),
+          (source) => filter(source, ([key, entry]) => !entry.cancel),
           (source) => filter(source, ([key, entry]) => !entry.cancel),
```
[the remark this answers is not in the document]
achingbrain: Have removed.

packages/bitswap/test/session.spec.ts
```diff
@@ -1,75 +1,177 @@
...
 
-    components.network.findProviders.returns(async function * () {
-      yield {
-        id: peerId,
-        multiaddrs: [],
-        protocols: ['']
+    components.network.findProviders.withArgs(cid).returns((async function * () {
+      for (const prov of providers) {
+        yield prov
+        await delay(100)
```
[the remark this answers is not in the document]
achingbrain: It needs a bit of a delay to simulate letting the query continue after we start to resolve blocks.  The slower it is, the more likely it'll work reliably on CI machines.

packages/block-brokers/src/bitswap.ts
```diff
@@ -72,16 +72,16 @@ class BitswapBlockBroker implements BlockBroker<BitswapWantBlockProgressEvents,
...
       },
```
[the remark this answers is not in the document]
achingbrain: Answered elsewhere: https://github.com/ipfs/helia/pull/495#discussion_r1563938765

packages/block-brokers/src/trustless-gateway/session.ts
achingbrain: The `filterMultiaddrs` function takes care of removing http address when `allowInsecure` is `false` and local addresses when `allowLocal` is false so the list should only contain addresses valid according to the config - what extra filtering were you thinking?

packages/utils/src/abstract-session.ts
achingbrain: Answered elsewhere: https://github.com/ipfs/helia/pull/495#discussion_r1563942408

> linter should be catching this and required a prefixed void

`void` is only required if it's a floating promise.  There's a `.catch(...)` here so it's not floating, so no `void` required by the linter.

packages/utils/src/abstract-session.ts
achingbrain: > we should probably normalize logging messages

Fair enough, this was mostly copied from the bitswap session impl.

> Also, should we add announce here at the abstract session level?

Maybe in the future but currently only bitswap needs `announce` and we should announce to any interested peer.

packages/utils/src/routing.ts
achingbrain: It's basically `p-queue` but more flexible..

packages/utils/src/abstract-session.ts
achingbrain: Done!

packages/block-brokers/src/trustless-gateway/session.ts
SgtPooki: I was mostly only thinking of filtering on allowInsecure and allowLocal because [0] might later get filtered out even though there are potentially more valid maddrs (e.g. at index 4)

packages/block-brokers/src/bitswap.ts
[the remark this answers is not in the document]
SgtPooki: It sounds like the `this.bitswap` instance will handle this for us then, and we don't need to use a scoped-down `session.notify`.

thanks for the explainer

PR #503, opened 2024-04-18
fix: prevent duplicate trustless-gateway reqs

SgtPooki:
- chore: add @types/polka
- feat: getStats() returns trustless gateway stats

## Title

fix: prevent duplicate trustless-gateway reqs

## Description

Fixes https://github.com/ipfs-shipyard/service-worker-gateway/issues/104

This PR fixes issues brought up in service-worker-gateway where sub-resources end up causing multiple requests to a trustless gateway for the root CID.

## Notes & open questions

I feel like the blockstore should handle this and ensure requests that depend on a root CID are deduped, but since trustless-gateways may be used outside of the context of blockstores, I believe we still need this improvement.

## Change checklist

### DIFF
--- a/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
+++ b/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
@@ -7,6 +7,7 @@ export interface TrustlessGatewayStats {
   errors: number
   invalidBlocks: number
   successes: number
+  pendingResponses?: number
 }
 
 /**
@@ -170,7 +171,8 @@ export class TrustlessGateway {
       attempts: this.#attempts,
       errors: this.#errors,
       invalidBlocks: this.#invalidBlocks,
-      successes: this.#successes
+      successes: this.#successes,
+      pendingResponses: this.#pendingResponses.size
     }
   }
 }
--- a/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
+++ b/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
@@ -1,3 +1,4 @@
+import { base64 } from 'multiformats/bases/base64'
 import type { ComponentLogger, Logger } from '@libp2p/interface'
 import type { CID } from 'multiformats/cid'
 
@@ -59,6 +60,20 @@ export class TrustlessGateway {
   }
 
   /**
+   * This function returns a unique string for the multihash.bytes of the CID.
+   *
+   * Some useful resources for why this is needed can be found using the links below:
+   *
+   * - https://github.com/ipfs/helia/pull/503#discussion_r1572451331
+   * - https://github.com/ipfs/kubo/issues/6815
+   * - https://www.notion.so/pl-strflt/Handling-ambiguity-around-CIDs-9d5e14f6516f438980b01ef188efe15d#d9d45cd1ed8b4d349b96285de4aed5ab
+   */
+  #uniqueBlockId (cid: CID): string {
+    const multihashBytes = cid.multihash.bytes
+    return base64.encode(multihashBytes)
+  }
+
+  /**
    * Fetch a raw block from `this.url` following the specification defined at
    * https://specs.ipfs.tech/http-gateways/trustless-gateway/
    */
@@ -74,8 +89,9 @@ export class TrustlessGateway {
       throw new Error(`Signal to fetch raw block for CID ${cid} from gateway ${this.url} was aborted prior to fetch`)
     }
 
+    const blockId = this.#uniqueBlockId(cid)
     try {
-      let pendingResponse: Promise<Uint8Array> | undefined = this.#pendingResponses.get(gwUrl.toString())
+      let pendingResponse: Promise<Uint8Array> | undefined = this.#pendingResponses.get(blockId)
       if (pendingResponse == null) {
         this.#attempts++
         pendingResponse = fetch(gwUrl.toString(), {
@@ -93,7 +109,7 @@ export class TrustlessGateway {
           this.#successes++
           return new Uint8Array(await res.arrayBuffer())
         })
-        this.#pendingResponses.set(gwUrl.toString(), pendingResponse)
+        this.#pendingResponses.set(blockId, pendingResponse)
       }
       return await pendingResponse
     } catch (cause) {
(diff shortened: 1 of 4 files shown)


SgtPooki (commented):
self review of the code so far that should fail CI test because fix isn't pushed yet.

SgtPooki:
note the failure which https://github.com/ipfs/helia/actions/runs/8745897909/job/24001821684?pr=503#step:5:793 [`97a05e4` (#503)](https://github.com/ipfs/helia/pull/503/commits/97a05e468f5dfbe642cecc6fa17fb370574c6be5) should fix

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -61,31 +75,34 @@ export class TrustlessGateway {
...
-      this.#successes++
-      return new Uint8Array(await res.arrayBuffer())
+      return await pendingResponse
     } catch (cause) {
       // @ts-expect-error - TS thinks signal?.aborted can only be false now
       // because it was checked for true above.
       if (signal?.aborted === true) {
         throw new Error(`fetching raw block for CID ${cid} from gateway ${this.url} was aborted`)
       }
+      this.log.error('failed to get block for %c from %s', cid, gwUrl, cause)
```
SgtPooki, line 105: idk if we didn't log errors here for a specific reason, but it feels like we should be logging the `cause` error so we can get details if needed.

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -61,31 +75,34 @@ export class TrustlessGateway {
...
+          },
+          cache: 'force-cache'
+        }).then(async (res) => {
+          this.log('GET %s %d', gwUrl, res.status)
+          if (!res.ok) {
+            this.#errors++
+            throw new Error(`unable to fetch raw block for CID ${cid} from gateway ${this.url}`)
+          }
+          this.#successes++
+          return new Uint8Array(await res.arrayBuffer())
```
SgtPooki, line 94: we need to fully encapsulate the request and response processing to ensure we're not attempting to consume the body twice. open to ideas of better methods.

SgtPooki (commented):
self review of code entirety

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
2color: Since you immediately handle the promise and don't expose `Response` object, where's the risk of the body getting consumed twice?

SgtPooki: There isnt anymore. In my first iteration i was only saving resp promise (without arrayBuffer handling) to the map.

This was more of a note about why we're doing promise.then without any awaits here

SgtPooki:
One thing I thought of after stepping away last night: we should probably use finally to remove entry from map

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -61,31 +75,34 @@ export class TrustlessGateway {
...
-        },
-        cache: 'force-cache'
-      })
-
-      this.log('GET %s %d', gwUrl, res.status)
-
-      if (!res.ok) {
-        this.#errors++
-        throw new Error(`unable to fetch raw block for CID ${cid} from gateway ${this.url}`)
+      let pendingResponse: Promise<Uint8Array> | undefined = this.#pendingResponses.get(gwUrl.toString())
```
achingbrain, line 78: We should dedupe responses based on a canonical stringification of the multihash of the CID (base64 or something) rather than then requested URL, because you can request two CIDs with different versions/codecs but they'd resolve to the same block.

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
achingbrain: It doesn't hurt, but in the places this method is invoked from the caller will also log the error so you can end up with duplicated stacks in the logs which can be very noisy.

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -37,6 +44,13 @@ export class TrustlessGateway {
    */
   #successes = 0
 
+  /**
+   * A map of pending responses for this gateway. This is used to ensure that
+   * only one request per CID is made to a given gateway at a time, and that we
+   * don't make multiple in-flight requests for the same CID to the same gateway.
+   */
+  #pendingResponses = new Map<string, Promise<Uint8Array>>()
```
achingbrain, line 52: I'm not sure anything is ever being removed from this map - there doesn't appear to be a `this.#pendingResponses.delete` anywhere in this PR so it will grow until the process falls over.

achingbrain (changes requested):
Seems reasonable, comments inline but please ensure things are removed from the pending responses map as it'll cause a memory leak as implemented.

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
SgtPooki: Good eye. I had it in there and didnt realize i lost it when refactoring to fix dupe body reads

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
SgtPooki: I put a codepen together to try and play with how this is done, and had some helpful tips from Adin to get me going

https://codepen.io/SgtPooki/pen/XWQygKB?editors=1111

it would be interesting to see where this falls down.

Also, is it just me that thinks it would be useful to export a blessed "are these CIDs referencing the same bytes" toString() function from multiformats?

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
SgtPooki: will remove

packages/block-brokers/test/trustless-gateway.spec.ts
```diff
@@ -190,4 +190,33 @@ describe('trustless-gateway-block-broker', () => {
 
     await expect(sessionBlockstore?.retrieve?.(blocks[0].cid)).to.eventually.deep.equal(blocks[0].block)
   })
+
+  it('does not trigger new network requests if the same cid request is in-flight', async function () {
```
SgtPooki, line 194: this test was renamed in the last iteration from "[...] same block [...]" to "[...] same cid [...]" because we're not actually testing the uniqueBlock functionality here where we would prevent requests given separate CIDs if they're referencing the same bytes.

SgtPooki (commented):
self review after addressing all comments

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
SgtPooki: Leaving this PR open to make sure you take a peek at this before I merge

achingbrain (approved):

PR #512, opened 2024-04-22
feat: add metrics property to helia interface

achingbrain:
To allow collecting metrics about arbitrary parts of the Helia stack, add an optional `.metrics` property for stat collection.

This can be used with implementations such as `@libp2p/prometheus-metrics` and/or others.

## Change checklist

### DIFF
--- a/packages/bitswap/src/network.ts
+++ b/packages/bitswap/src/network.ts
@@ -15,7 +15,7 @@ import type { WantOptions } from './bitswap.js'
 import type { MultihashHasherLoader } from './index.js'
 import type { Block } from './pb/message.js'
 import type { Provider, Routing } from '@helia/interface/routing'
-import type { Libp2p, AbortOptions, Connection, PeerId, IncomingStreamData, Topology, ComponentLogger, IdentifyResult, Counter } from '@libp2p/interface'
+import type { Libp2p, AbortOptions, Connection, PeerId, IncomingStreamData, Topology, ComponentLogger, IdentifyResult, Counter, Metrics } from '@libp2p/interface'
 import type { Logger } from '@libp2p/logger'
 import type { CID } from 'multiformats/cid'
 import type { ProgressEvent, ProgressOptions } from 'progress-events'
@@ -49,6 +49,7 @@ export interface NetworkComponents {
   routing: Routing
   logger: ComponentLogger
   libp2p: Libp2p
+  metrics?: Metrics
 }
 
 export interface BitswapMessageEventDetail {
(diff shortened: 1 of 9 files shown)


packages/interface/src/index.ts
```diff
@@ -74,6 +74,12 @@ export interface Helia {
    */
   dns: DNS
 
+  /**
+   * A metrics object that can be used to collected arbitrary stats about node
+   * usage.
```
SgtPooki, line 79: It looks like we use libp2p metrics config object directly. Should we link to https://github.com/libp2p/js-libp2p/blob/main/doc/CONFIGURATION.md#configuring-metrics ?

packages/utils/src/index.ts
```diff
@@ -131,6 +131,12 @@ export interface HeliaInit {
    * An optional DNS implementation used to perform queries for DNS records.
    */
   dns?: DNS
+
+  /**
+   * A metrics object that can be used to collected arbitrary stats about node
+   * usage.
```
SgtPooki, line 137: ditto on linking to https://github.com/libp2p/js-libp2p/blob/main/doc/CONFIGURATION.md#configuring-metrics ?

packages/bitswap/src/network.ts
```diff
@@ -101,13 +102,13 @@ export class Network extends TypedEventEmitter<NetworkEvents> {
     this.maxIncomingMessageSize = init.maxIncomingMessageSize ?? DEFAULT_MAX_OUTGOING_MESSAGE_SIZE
     this.maxOutgoingMessageSize = init.maxOutgoingMessageSize ?? init.maxIncomingMessageSize ?? DEFAULT_MAX_INCOMING_MESSAGE_SIZE
     this.metrics = {
-      blocksSent: components.libp2p.metrics?.registerCounter('helia_bitswap_sent_blocks_total'),
-      dataSent: components.libp2p.metrics?.registerCounter('helia_bitswap_sent_data_bytes_total')
+      blocksSent: components.metrics?.registerCounter('helia_bitswap_sent_blocks_total'),
+      dataSent: components.metrics?.registerCounter('helia_bitswap_sent_data_bytes_total')
```
SgtPooki, line 106: we should update readme with a list of metrics that we add when metrics are enabled.

packages/bitswap/src/stats.ts
```diff
@@ -11,10 +12,10 @@ export class Stats {
...
 
   constructor (components: StatsComponents) {
-    this.blocksReceived = components.libp2p.metrics?.registerMetricGroup('helia_bitswap_received_blocks')
-    this.duplicateBlocksReceived = components.libp2p.metrics?.registerMetricGroup('helia_bitswap_duplicate_received_blocks')
-    this.dataReceived = components.libp2p.metrics?.registerMetricGroup('helia_bitswap_data_received_bytes')
-    this.duplicateDataReceived = components.libp2p.metrics?.registerMetricGroup('helia_bitswap_duplicate_data_received_bytes')
+    this.blocksReceived = components.metrics?.registerMetricGroup('helia_bitswap_received_blocks')
+    this.duplicateBlocksReceived = components.metrics?.registerMetricGroup('helia_bitswap_duplicate_received_blocks')
+    this.dataReceived = components.metrics?.registerMetricGroup('helia_bitswap_data_received_bytes')
+    this.duplicateDataReceived = components.metrics?.registerMetricGroup('helia_bitswap_duplicate_data_received_bytes')
```
SgtPooki, line 18: we should update readme with a list of metrics that we add when metrics are enabled.

SgtPooki (approved):
approved, but we should update readme with explainers on metrics collected, or a list of them at a minimum

packages/interface/src/index.ts
achingbrain: Those docs need some updating, they're not correct any more.

packages/bitswap/src/network.ts
achingbrain: That would be good, though prone to being out of date very quickly.  It'd be nice if we could automate this somehow.

PR #514, opened 2024-04-22
fix: log peer id as string not object

achingbrain:
So we aren't logging buffers etc, log the peer id as a string when lookup fails.

## Change checklist

### DIFF
--- a/packages/utils/src/routing.ts
+++ b/packages/utils/src/routing.ts
@@ -89,7 +89,7 @@ export class Routing implements RoutingInterface, Startable {
 
             return provider
           } catch (err) {
-            this.log.error('could not load multiaddrs for peer', peer.id, err)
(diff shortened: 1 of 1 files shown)


SgtPooki (approved):

packages/utils/src/routing.ts
```diff
@@ -89,15 +89,15 @@ export class Routing implements RoutingInterface, Startable {
 
             return provider
           } catch (err) {
-            this.log.error('could not load multiaddrs for peer', peer.id, err)
+            this.log.error('could not load multiaddrs for peer %p', peer.id, err)
```
2color, line 92: What is the `%p` substitution?

SgtPooki: Custom libp2p logger formatter. Libp2p has a number of these for multiaddr, cid, peerid, etc..

PR #515, opened 2024-04-23
feat: add static http gateway routing

achingbrain:
Adds a routing implementation that returns a static list of gateways as providers for CIDs.

It's expected that these gateways are able to fetch content on our behalf.

Depends on:

- [x] https://github.com/libp2p/js-libp2p/pull/2496
- [x] https://github.com/ipfs/helia/pull/516

## Change checklist

### DIFF
--- a/packages/routers/src/http-gateway-routing.ts
+++ b/packages/routers/src/http-gateway-routing.ts
@@ -4,7 +4,7 @@ import { identity } from 'multiformats/hashes/identity'
 import { fromString as uint8ArrayFromString } from 'uint8arrays/from-string'
 import { toString as uint8ArrayToString } from 'uint8arrays/to-string'
 import type { Provider, Routing, RoutingOptions } from '@helia/interface'
-import type { PeerId, PeerInfo } from '@libp2p/interface'
+import { peerIdSymbol, type PeerId, type PeerInfo } from '@libp2p/interface'
 import type { MultihashDigest, Version } from 'multiformats'
 
 export interface HTTPGatwayRouterInit {
@@ -30,6 +30,8 @@ class URLPeerId {
   [inspect] (): string {
     return `PeerId(${this.url})`
   }
+
+  readonly [peerIdSymbol] = true
 
   toString (): string {
     return this.toCID().toString()
--- a/packages/routers/src/http-gateway-routing.ts
+++ b/packages/routers/src/http-gateway-routing.ts
@@ -1,0 +1,90 @@
+import { uriToMultiaddr } from '@multiformats/uri-to-multiaddr'
+import { CID } from 'multiformats/cid'
+import { identity } from 'multiformats/hashes/identity'
+import { fromString as uint8ArrayFromString } from 'uint8arrays/from-string'
+import { toString as uint8ArrayToString } from 'uint8arrays/to-string'
+import type { Provider, Routing, RoutingOptions } from '@helia/interface'
+import type { PeerId, PeerInfo } from '@libp2p/interface'
+import type { MultihashDigest, Version } from 'multiformats'
+
+export interface HTTPGatwayRouterInit {
+  gateways?: Array<URL | string>
+}
+
+// these values are from https://github.com/multiformats/multicodec/blob/master/table.csv
+const LIBP2P_KEY_CODE = 0x72
+const inspect = Symbol.for('nodejs.util.inspect.custom')
+
+class URLPeerId {
+  readonly type = 'url'
+  readonly multihash: MultihashDigest
+  readonly privateKey?: Uint8Array
+  readonly publicKey?: Uint8Array
+  readonly url: string
+
+  constructor (url: URL) {
+    this.url = url.toString()
+    this.multihash = identity.digest(uint8ArrayFromString(this.url))
+  }
+
+  [inspect] (): string {
+    return `PeerId(${this.url})`
+  }
+
+  toString (): string {
+    return this.toCID().toString()
+  }
+
+  toCID (): CID {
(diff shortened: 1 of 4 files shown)


packages/routers/src/http-gateway-routing.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export interface HTTPGatwayRouterInit {
+  gateways?: Array<URL | string>
+}
+
+// these values are from https://github.com/multiformats/multicodec/blob/master/table.csv
+const LIBP2P_KEY_CODE = 0x72
+const inspect = Symbol.for('nodejs.util.inspect.custom')
+
+class URLPeerId {
```
2color, line 18: Is this supposed to implement one of the PeerID types?

packages/routers/src/http-gateway-routing.ts
```diff
@@ -0,0 +1,92 @@
...
+import type { Provider, Routing, RoutingOptions } from '@helia/interface'
+import { peerIdSymbol, type PeerId, type PeerInfo } from '@libp2p/interface'
+import type { MultihashDigest, Version } from 'multiformats'
+
+export interface HTTPGatwayRouterInit {
+  gateways?: Array<URL | string>
+}
+
+// these values are from https://github.com/multiformats/multicodec/blob/master/table.csv
+const LIBP2P_KEY_CODE = 0x72
```
2color, line 15: Does it make sense to use the [code for `transport-ipfs-gateway-http` instead](https://github.com/multiformats/multicodec/blob/master/table.csv#L149)?
 
```suggestion
const TRANSPORT_IPFS_GATEWAY_HTTP_CODE = 0x0920
```

packages/routers/src/http-gateway-routing.ts
```diff
@@ -0,0 +1,92 @@
...
+class URLPeerId {
+  readonly type = 'url'
+  readonly multihash: MultihashDigest
+  readonly privateKey?: Uint8Array
+  readonly publicKey?: Uint8Array
+  readonly url: string
+
+  constructor (url: URL) {
+    this.url = url.toString()
+    this.multihash = identity.digest(uint8ArrayFromString(this.url))
```
2color, line 27: What's the reasoning for using the identity hash rather than an actual hash which will be of consistent size (unlike gw urls which vary in length) making the `toCID` return consistent length CIDs?

packages/routers/src/http-gateway-routing.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  readonly [peerIdSymbol] = true
+
+  toString (): string {
+    return this.toCID().toString()
+  }
+
+  toCID (): CID {
+    return CID.createV1(LIBP2P_KEY_CODE, this.multihash)
```
2color, line 41: Related https://github.com/ipfs/helia/pull/515/files#r1577825895

2color:
I'm curious, how do you plan on using the HTTPGatwayRouter once merged?

packages/routers/src/http-gateway-routing.ts
achingbrain: An actual hash would require async to create, which we'd have to handle somewhere whereas this is sync and sync is simple/fast.

That said, it's just to have something unique that lets us compare peer identifiers - is there any benefit to having consistent length CIDs?

achingbrain:
> how do you plan on using the HTTPGatwayRouter once merged?

It'll be configured like any other router and will let http gateway sessions fall back to a default set of gateways.

We'll also be able to switch the gateway block broker over to use the helia routing which will let it find non-preconfigured gateways but also fall back to the default ones.

packages/routers/src/http-gateway-routing.ts
achingbrain: For completeness it can implement `PeerId` though as of until a few minutes ago `PeerId` was a union type of `Ed`, `RSA` and `sepk` keys, of which this is none of them.

`URLPeerId` already has type overlap with `PeerId` so it doesn't need to explicitly implement the interface but it does make the intention clearer.

Anyway it's an interface now that can be implemented by arbitrary identifier types so it can be switched over.

SgtPooki (approved):
looks straightforward to me but i'm not intimately familiar with PeerId class we're implementing. 

i think sync multihash is fine, toCID should be unique per URL.. ship it

SgtPooki:
> I'm curious, how do you plan on using the HTTPGatwayRouter once merged?

Some context for @2color from chat I had with @achingbrain when talking about issues with trustless-gateways related to testing out the changes in https://github.com/ipfs/helia-verified-fetch/pull/50

> Basically I’m not sure we’re ready for HTTP Gateway sessions yet.
> To that end I’ve put this PR together that adds a routing implementation that just returns a static list of gateways - https://github.com/ipfs/helia/pull/515
> This way when we get better HTTP results via the delegated lookup in the future we can just turn off the static list.

Essentially, we need some way to reduce the breadth of content-fetching requests (via sessions), and trustless gateways returned from delegated routing providers are not consistently reliable and frequently not available at all.

When I was looking into it, I proposed having trustless gateway sessions fall back to the provided trustless-gateways passed to blockBrokers. IIUC, this PR enables a version of that fallback in the defined "routing" language we've recently moved to in helia.

Alex to correct me here if I'm wrong or too imprecise.

achingbrain:
That's it, all good.

packages/routers/src/http-gateway-routing.ts
2color: Got it! 

> is there any benefit to having consistent length CIDs?
Not that I can think of.

2color:
Nice! Cool stuff

PR #516, opened 2024-04-23
fix: do not append peer ids to provider multiaddrs

achingbrain:
When a routing returns a set of multiaddrs, honour them as returned instead of appending the peer id to them.

## Change checklist

### DIFF
--- a/packages/utils/src/routing.ts
+++ b/packages/utils/src/routing.ts
@@ -64,14 +64,6 @@ export class Routing implements RoutingInterface, Startable {
         continue
       }
 
-      peer.multiaddrs = peer.multiaddrs.map(ma => {
-        if (ma.getPeerId() != null) {
-          return ma
-        }
-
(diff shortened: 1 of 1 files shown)


packages/utils/src/routing.ts
```diff
@@ -64,14 +64,6 @@ export class Routing implements RoutingInterface, Startable {
         continue
       }
 
-      peer.multiaddrs = peer.multiaddrs.map(ma => {
-        if (ma.getPeerId() != null) {
-          return ma
-        }
-
-        return ma.encapsulate(`/p2p/${peer.id}`)
```
2color, line 72: Why were we doing this in the first place?

achingbrain: Good question, I don't think it's necessary since we only care when we try to dial a multiaddr, and libp2p appends it if it's not there at that point.

PR #519, opened 2024-04-25
fix: http blockbroker loads gateways from routing

achingbrain:
Use the new http gateway block broker to load gateways from the routing.

If we can find http gateways for a block we now use it, falling back to the default list of gateways if the routing is slow or it gives providers that fail to supply the block.

One change here is that unless you are using sessions, trustless gateways become disposable so we can't sort them by reliability any more.

I'm not sure if we have any stats on the real-world impact of being able to do this?  It may be more relevant for sessions anyway, I think.

## Change checklist


packages/block-brokers/src/trustless-gateway/broker.ts
```diff
@@ -29,55 +29,48 @@ export interface CreateTrustlessGatewaySessionOptions extends CreateSessionOptio
...
+
         try {
           await options.validateFn?.(block)
         } catch (err) {
           this.log.error('failed to validate block for %c from %s', cid, gateway.url, err)
-          gateway.incrementInvalidBlocks()
-
-          throw new Error(`Block for CID ${cid} from gateway ${gateway.url} failed validation`)
+          // try another gateway
+          continue
```
SgtPooki, line 61: do we want to ignore that some gateway provider that returned invalid blocks? could we maybe emit an event so a session could remove that gateway?

packages/routers/src/http-gateway-routing.ts
```diff
@@ -8,6 +8,17 @@ import type { Provider, Routing, RoutingOptions } from '@helia/interface'
 import type { PeerId, PeerInfo } from '@libp2p/interface'
 import type { MultihashDigest, Version } from 'multiformats'
 
+export const DEFAULT_TRUSTLESS_GATEWAYS = [
+  // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs-public-gateway-checker.on.fleek.co/
+  'https://trustless-gateway.link',
+
+  // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs-public-gateway-checker.on.fleek.co/
+  'https://cloudflare-ipfs.com',
```
SgtPooki, line 16: we might want to remove this one?

SgtPooki (approved):
in general, lgtm

packages/block-brokers/src/trustless-gateway/broker.ts
achingbrain: The session will remove the gateway if `queryProvider` throws, so as long as `getRawBlock` (or the block validator) throws it'll get evicted from the session.

packages/routers/src/http-gateway-routing.ts
```diff
@@ -8,6 +8,17 @@ import type { Provider, Routing, RoutingOptions } from '@helia/interface'
...
+
+  // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs-public-gateway-checker.on.fleek.co/
+  'https://4everland.io'
```
achingbrain, line 19:
```suggestion
  // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs-public-gateway-checker.on.fleek.co/
  'https://4everland.io'
```

packages/routers/src/http-gateway-routing.ts
achingbrain: It does appear to be down.  As does https://ipfs-public-gateway-checker.on.fleek.co

Le sigh.

SgtPooki (approved):

PR #538, opened 2024-05-16
feat: two explicit queues for session work

SgtPooki:
## Title

feat: two explicit queues for session work

## Description

This PR originally aimed to fix the infinite loop we hit when we were only getting a single provider (this only occurring in test envs that i'm aware of, but certainly possible elsewhere). However, it's now been updated to refactor the provider finding and querying logic so that it's a little more obvious what is happening.

## Notes & open questions

A few main things that change with this PR:

1. We are now explicitly kicking off the `findProviders` task when findProviders is done (`idle` event)
2. We are only querying providers when we get a `provider` event, or when an existing session is queried with a new cid (we will query any currently known and non-evicted providers). 
    - The latter was existing code that was always ran, but is no longer needed for non-child-item-retrieve events since we query on `provider` events now
4. We use the same `found < count` logic to determine if we found enough providers, or throw an error
5. On findProvider job error (`failure` event), we set a flag indicating that we failed to find more providers.
    - if queryProviders job is running, we do not throw this error, because queryProviderQueue `idle` will handle when triggered.
7. on queryProvider `idle` 
    - if we found the block: we don't start up any more work
    - if signal aborted, we reject the `retrieve` promise
    - if foundProviders errored, we reject the `retrieve` promise: new error that we're done querying providers and could not get the block
    - if findProviders queue is empty, we kick it off again (this case covers the child-item-t-retrieve event where findProvidersQueue idle would have already been emitted on root-CID query)

## Change checklist

### DIFF
--- a/packages/utils/src/abstract-session.ts
+++ b/packages/utils/src/abstract-session.ts
@@ -175,7 +175,7 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
     try {
       return await deferred.promise
     } finally {
-      this.log('finally block, cleaning up session')
+      this.log.trace('retrieve finished, cleaning up session')
       this.removeEventListener('provider', peerAddedToSessionListener)
       this.findProviderQueue.clear()
       this.queryProviderQueue.clear()
@@ -249,7 +249,6 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
    * @param options - AbortOptions
    */
   private async findProviders (cid: CID, count: number, options: AbortOptions): Promise<void> {
-    this.log('findProviders called')
     const deferred: DeferredPromise<void> = pDefer()
     let found = 0
 
@@ -266,10 +265,8 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
           }
 
           if (this.hasProvider(provider)) {
-            this.log('ignoring duplicate provider')
+            this.log.trace('ignoring duplicate provider')
             continue
-          } else {
-            this.log('provider is not a duplicate')
           }
 
           found++


packages/utils/test/abstract-session.spec.ts
```diff
@@ -233,4 +233,44 @@ describe('abstract-session', () => {
...
+    })
+
+    session.queryProvider.callsFake(async () => {
+      // always fails
+      throw new Error('Urk!')
+    })
+
+    await expect(session.retrieve(cid)).to.eventually.be.rejected()
+
+    expect(session.findNewProviders.callCount).to.equal(4)
```
SgtPooki, line 256: I don't think we should be calling this 4 times. At most, we should be calling it twice:

1. intialPeerSearchComplete
2. on idle, we call `this.findProviders(cid, this.minProviders, options)`

and then it should fail.

SgtPooki (commented):
self review

packages/utils/test/abstract-session.spec.ts
SgtPooki: thoughts @achingbrain ?

packages/utils/src/abstract-session.ts
```diff
@@ -95,8 +103,9 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
       deferred.resolve(evt.detail.result)
     })
     queue.addEventListener('idle', () => {
-      if (foundBlock || options.signal?.aborted === true) {
-        // we either found the block or the user gave up
+      if (foundBlock || options.signal?.aborted === true || this.noNewProviders) {
+        // we either found the block, the user gave up, or cannot find any more providers
+        this.log('session aborted')
```
SgtPooki, line 108:
```suggestion
        this.log('session finished')
```

packages/utils/src/abstract-session.ts
```diff
@@ -74,113 +90,139 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
...
       this.evict(evt.detail.job.options.provider)
     })
-    queue.addEventListener('success', (evt) => {
-      // peer has sent block, return it to the caller
+
+    this.queryProviderQueue.addEventListener('success', (event) => {
+      this.log.trace('queryProviderQueue success')
       foundBlock = true
-      deferred.resolve(evt.detail.result)
+      // this.findProviderQueue.clear()
```
SgtPooki:
```suggestion
```

packages/utils/src/abstract-session.ts
```diff
@@ -74,113 +90,139 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
...
-          if (options.signal?.aborted === true) {
-            // skip logging error if signal was aborted because abort can happen
-            // on success (e.g. another session found the block)
-            return
-          }
-
-          this.log.error('error retrieving session block for %c', cid, err)
-        })
+    try {
+      // this.intialPeerSearchComplete = this.findProviders(cid, this.minProviders, options)
```
SgtPooki:
```suggestion
```

packages/utils/src/abstract-session.ts
```diff
@@ -74,113 +90,135 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
...
+      }
+      if (['ERR_INSUFFICIENT_PROVIDERS_FOUND'].includes((evt.detail.error as CodeError).code)) {
+        this.log.error('insufficient providers found for %c', cid)
+        if (this.queryProviderQueue.running === 0) {
+          // only reject if we're not currently querying any providers
+          deferred.reject(evt.detail.error)
+        }
+      }
     })
-    queue.addEventListener('error', () => {})
```
SgtPooki, line 87: not sure if adding this empty error handler is necessary?

SgtPooki (commented):
self review. ready for @achingbrain

SgtPooki:
run benchmarks to see if regression or improvement.

PR #542, opened 2024-05-20
fix: check eviction filter for new providers

### LINKED ISSUE #501: bug: endless session routing
When looking at https://github.com/ipfs/helia-verified-fetch/pull/50, I ran into an issue: https://github.com/ipfs/helia-verified-fetch/pull/50#issuecomment-2062047400

It seems like a request for QmVP2ip92jQuMDezVSzQBWDqWFbp9nyCHNQSiciRauPLDg endlessly polls for providers of `QmVP2ip92jQuMDezVSzQBWDqWFbp9nyCHNQSiciRauPLDg` with the default delegated-ipfs.dev router. I think this is due to https://github.com/ipfs/helia/blob/13daa26ec9bc6c1a1552b7b56597945c5b13a87b/packages/utils/src/abstract-session.ts#L215-L254

something like `maxProviders++` is occurring (i.e. https://github.com/ipfs/helia/pull/495#discussion_r1563085044), but much less obviously

references:

* [ipfs/helia-verified-fetch@`647b6cb` (#55)](https://github.com/ipfs/helia-verified-fetch/pull/55/commits/647b6cb53c517c4c70cf196531d11574b934b5ca) was the first fix while investigating.

achingbrain:
When new providers are found, check the eviction filter in `isEvicted` rather than the current list of providers.

Fixes #501

## Change checklist


packages/utils/src/abstract-session.ts
```diff
@@ -188,7 +188,7 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
   }
 
   isEvicted (provider: Provider): boolean {
-    return this.providers.some(prov => this.equals(prov, provider))
+    return this.evictionFilter.has(this.toEvictionKey(provider))
```
SgtPooki, line 191: bah.. how did I miss that

SgtPooki (approved):

PR #543, opened 2024-05-20
test: hasProvider does not work for evicted providers

SgtPooki:
## Title

test: hasProvider does not work for evicted providers

## Description

I discovered that our evictionFilter (using the new BloomFilter) does not correctly return `true` inside `hasProvider` when the provider is evicted while looking into https://github.com/ipfs/helia/pull/538

## Change checklist


SgtPooki:
missed that this was already found and addressed with https://github.com/ipfs/helia/pull/542

PR #547, opened 2024-05-24
fix: do not depend on external domains in dnslink tests

lidel:
Dependency on third-party domains means  that over time CI will break, like it did in
https://github.com/ipfs/kubo/actions/runs/9214249430/job/25364989609?pr=10429#step:9:63

```
Error: queryCname ENODATA singularity.storage
```

The three domains I left are stable and under IPFS Foundation / Shipyard control.

Ideally, CI would not depend on internet DNS, and have static fixtures and custom localhost DNS server, but this PR should be good enough to stop breaking CI on other project's PRs like https://github.com/ipfs/kubo/pull/10429

## Change checklist

### DIFF
--- a/packages/interop/src/ipns-dnslink.spec.ts
+++ b/packages/interop/src/ipns-dnslink.spec.ts
@@ -7,17 +7,9 @@ import type { IPNS } from '@helia/ipns'
 import type { HeliaLibp2p } from 'helia'
 
 const TEST_DOMAINS: string[] = [
-  'ipfs.io',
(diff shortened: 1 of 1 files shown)


SgtPooki:
Fyi current helia-verified-fetch gwc PR (todo link) has some dnsLink stubbing code (IPFS_NS_MAP like functionality)

SgtPooki (approved):

2color (approved):

PR #560, opened 2024-06-18
fix: docs/example

kristian-nygaard-jensen:
fs.add is not a function
"Invalid key type" key type must be specified

https://github.com/libp2p/js-libp2p/blob/main/packages/keychain/src/keychain.ts#L157

## Description

Two function calls need to be modified for the examples to work.

## Notes & open questions

I will try to write tests that prove the issues

## Change checklist

### DIFF
--- a/packages/ipns/README.md
+++ b/packages/ipns/README.md
@@ -45,12 +45,12 @@ const helia = await createHelia()
 const name = ipns(helia)
 
 // create a public key to publish as an IPNS name
-const keyInfo = await helia.libp2p.services.keychain.createKey('my-key')
+const keyInfo = await helia.libp2p.services.keychain.createKey('my-key', 'RSA', 4096)
 const peerId = await helia.libp2p.services.keychain.exportPeerId(keyInfo.name)
 
 // store some data to publish
 const fs = unixfs(helia)
-const cid = await fs.add(Uint8Array.from([0, 1, 2, 3, 4]))
+const cid = await fs.addBytes(Uint8Array.from([0, 1, 2, 3, 4]))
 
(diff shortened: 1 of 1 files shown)


packages/ipns/README.md
```diff
@@ -45,12 +45,12 @@ const helia = await createHelia()
 const name = ipns(helia)
 
 // create a public key to publish as an IPNS name
-const keyInfo = await helia.libp2p.services.keychain.createKey('my-key')
+const keyInfo = await helia.libp2p.services.keychain.createKey('my-key', 'RSA', 4096)
```
achingbrain, line 48: Ed keys are preferred over RSA for new code, it'd be better if the docs showed their use instead:

```suggestion
const keyInfo = await helia.libp2p.services.keychain.createKey('my-key', 'Ed25519')
```

achingbrain (changes requested):
Thanks for opening this - can you please swap refs to RSA keys to use Ed25519 keys instead?

packages/ipns/README.md
kristian-nygaard-jensen: Sure

achingbrain: TBH the keychain should just default to Ed keys..

SgtPooki (changes requested):
since the changes are not made to the index.ts files, they will be overwritten. can you make the modifications to the index.ts files so our doc gen updates the readme.md?

SgtPooki:
FYI we have doc-check enabled in all repos now and it catches any invalid examples. It seems like `packages/ipns` is passing on main when using node@22. closing this

PR #565, opened 2024-06-19
docs: fix grammar - it's -> its

jaller94:
## Title

Fix grammar: it's -> its

## Description

A few places use "it's" (= it is), when they should use the possessive "its" (the dog and its ball).

## Change checklist

### DIFF
--- a/packages/bitswap/README.md
+++ b/packages/bitswap/README.md
@@ -34,7 +34,7 @@ $ npm i @helia/bitswap
 
 ## Browser `<script>` tag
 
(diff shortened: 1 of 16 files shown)


achingbrain (approved):

SgtPooki (changes requested):
Thanks so much for the contribution! FYI the documentation in these readme.md files are generated from the package documentation jsdoc in the index.ts files for each of the packages in the monorepo

achingbrain:
The part of the README in question is generated by aegir so I think we're all good here - ref: https://github.com/ipfs/aegir/pull/1566

PR #566, opened 2024-06-21
fix: respect trustless gateway options for sessions

2color:
## Description

Fixes an [insidious bug](https://github.com/ipfs/service-worker-gateway/pull/303#issuecomment-2180905836), whereby the `allowInsecure` and `allowLocal` options of the trustless gateway broker were not respected for sessions. 

This PR ensures that they are correctly passed down to the TrustlessGatewaySession class.

## Change checklist

### DIFF
--- a/packages/helia/src/utils/libp2p-defaults.browser.ts
+++ b/packages/helia/src/utils/libp2p-defaults.browser.ts
@@ -32,7 +32,7 @@ export interface DefaultLibp2pServices extends ServiceMap {
   ping: PingService
 }
 
-export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOptions<DefaultLibp2pServices> {
+export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOptions<DefaultLibp2pServices> & Required<Pick<Libp2pOptions<DefaultLibp2pServices>, 'services'>> {
   const agentVersion = `${name}/${version} ${libp2pInfo.name}/${libp2pInfo.version} UserAgent=${globalThis.navigator.userAgent}`
 
   return {
--- a/packages/helia/src/utils/libp2p-defaults.browser.ts
+++ b/packages/helia/src/utils/libp2p-defaults.browser.ts
@@ -19,9 +19,10 @@ import * as libp2pInfo from 'libp2p/version'
 import { name, version } from '../version.js'
 import { bootstrapConfig } from './bootstrappers.js'
 import type { Libp2pDefaultsOptions } from './libp2p.js'
+import type { ServiceMap } from '@libp2p/interface'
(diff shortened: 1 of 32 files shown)


packages/helia/src/utils/libp2p-defaults.browser.ts
```diff
@@ -19,9 +19,10 @@ import * as libp2pInfo from 'libp2p/version'
 import { name, version } from '../version.js'
 import { bootstrapConfig } from './bootstrappers.js'
 import type { Libp2pDefaultsOptions } from './libp2p.js'
+import type { ServiceMap } from '@libp2p/interface'
 import type { Libp2pOptions } from 'libp2p'
 
-export interface DefaultLibp2pServices extends Record<string, unknown> {
+export interface DefaultLibp2pServices extends ServiceMap {
```
SgtPooki, line 25: I'm not sure about the best way to extend libp2p service map type. do we need to pass through a generic here?

packages/helia/src/utils/libp2p-defaults.ts
```diff
@@ -21,9 +21,10 @@ import * as libp2pInfo from 'libp2p/version'
 import { name, version } from '../version.js'
 import { bootstrapConfig } from './bootstrappers.js'
 import type { Libp2pDefaultsOptions } from './libp2p.js'
+import type { ServiceMap } from '@libp2p/interface'
 import type { Libp2pOptions } from 'libp2p'
 
-export interface DefaultLibp2pServices extends Record<string, unknown> {
+export interface DefaultLibp2pServices extends ServiceMap {
```
SgtPooki, line 27: ditto

SgtPooki (approved):
I think it would be good to get this in

packages/helia/src/utils/libp2p-defaults.browser.ts
2color: I don't remember why I did this to be honest, as this PR was blocked because of a type error related to Libp2p which was fixed in https://github.com/ipfs/helia/pull/566/commits/d123032976089a1eebf7ccbc64f19bc370c8eb23. 

Will need to take a closer look, but maybe @achingbrain can weigh in.

achingbrain: It doesn't look like the changes to the generics here are necessary any more, they can be reverted.

achingbrain (approved):

packages/helia/src/utils/libp2p-defaults.browser.ts
achingbrain: It's weird that they were necessary in the first place since the type `ServiceMap` is `Record<string, unknown>` itself so there's 100% overlap 🤷 .

Issue #564, opened 2024-06-19
`allowInsecure: false` should respect Secure Contexts


lidel:
## Problem

Setting `allowInsecure: false` should not be blindly blocking `http://`, it should not block requests to `http://localhost`  and `http://*.localhost`  because these are valid Secure Contexts. 

This bug blocks users from using their own local gateway (ipfs desktop, kubo, rainbow).

## Solution

Correctly recognize `http://localhost[:port]`  and `http://*.localhost[:port]` as secure contexts.

> [!IMPORTANT]
> Only `localhost` label is marked as Secure Context, URLs with loopback `127.0.0.1` IPs are not.

Ref. 
- https://github.com/ipfs/helia/blob/74ccc92793a6d0bb4bee714d9fe4fa4183aa4ee8/packages/block-brokers/src/trustless-gateway/utils.ts#L12
- TLDR: https://github.com/ipfs/in-web-browsers/issues/109#issuecomment-1344368343
- W3C: https://w3c.github.io/webappsec-secure-contexts/#localhost
- IETF: https://datatracker.ietf.org/doc/html/draft-west-let-localhost-be-localhost

2color:
According to https://www.w3.org/TR/secure-contexts/#is-origin-trustworthy is seems that `127.0.0.1` is potentially trustworthy:

> If origin’s [host](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-host) matches one of the CIDR notations 127.0.0.0/8 or ::1/128 [[RFC4632]](https://w3c.github.io/webappsec-secure-contexts/#biblio-rfc4632), return "Potentially Trustworthy".

SgtPooki:
For anyone interested in taking this task on (e.g. Luca from 2024-08-01 Helia WG): 

We will want to modify https://github.com/ipfs/helia/blob/74ccc92793a6d0bb4bee714d9fe4fa4183aa4ee8/packages/block-brokers/src/trustless-gateway/utils.ts#L10C17-L31 in the following ways:

1. Modify the filter so that when `allowInsecure=false`, a multiaddr with "127.0.0.1" as the IP still returns true. add a test
2. Modify the filter so that when `allowInsecure=false`, a multiaddr with "localhost" as the domain still returns true. add a test
3. Modify the filter so that when `allowInsecure=false`, a multiaddr with a parent domain of "localhost" as the domain still returns true (e.g. `example.localhost`, `foobar.localhost`, `*.localhost`). add a test

PR #579, opened 2024-08-08
fix: Adjust filtering logic for secure contexts; improve tests

acul71:
## Title
fix: Update filtering logic to respect secure contexts and enhance related tests

## Description
This PR addresses issue #564 by modifying the multiaddr filtering logic in the trustless-gateway module to ensure that allowInsecure: false respects Secure Contexts.

## Changes Made:
IP Address Filtering:
Modified the filter to ensure that when allowInsecure is false, a multiaddr with "127.0.0.1" as the IP address will still be considered valid. This addresses the need to handle secure contexts properly even when allowInsecure is disabled.
Domain Filtering:
Adjusted the filter to ensure that when allowInsecure is false, a multiaddr with "localhost" as the domain will be treated as valid.
Updated the filter to handle domains with a parent domain of "localhost", such as example.localhost, foobar.localhost, and *.localhost.
```javascript
// When allowInsecure is false and allowLocal is true, allow multiaddrs with "127.0.0.1", "localhost", or any subdomain ending with ".localhost"
    if (!allowInsecure && allowLocal) {
      if (ma.toOptions().host === '127.0.0.1' || ma.toOptions().host === 'localhost' || ma.toOptions().host.endsWith('.localhost'))
        return true
    }
```

## Tests Added:
Added tests to verify that the changes work as intended:
IP Address Test: Ensures that "127.0.0.1" is treated correctly when allowInsecure is false.
Domain Test: Verifies that "localhost" and parent domains of "localhost" are handled correctly when allowInsecure is false.
Related: [GitHub Issue #564](https://github.com/ipfs/helia/issues/564)
```javascript

  it('filterNonHTTPMultiaddrs allows 127.0.0.1 when allowInsecure=false', async function () {
    const localMaddr = uriToMultiaddr('http://127.0.0.1')

    const filtered = filterNonHTTPMultiaddrs([localMaddr], false, true)

    expect(filtered.length).to.deep.equal(1)
  })

  it('filterNonHTTPMultiaddrs allows localhost when allowInsecure=false', async function () {
    const localMaddr = uriToMultiaddr('http://localhost')

    const filtered = filterNonHTTPMultiaddrs([localMaddr], false, true)

    expect(filtered.length).to.deep.equal(1)
  })

  it('filterNonHTTPMultiaddrs allows *.localhost when allowInsecure=false', async function () {
    const localMaddr = uriToMultiaddr('http://example.localhost')

    const filtered = filterNonHTTPMultiaddrs([localMaddr], false, true)

    expect(filtered.length).to.deep.equal(1)
  })
```

## Notes & open questions
None at the moment. The changes have been tested and are working as expected.

## Change checklist

### DIFF
--- a/packages/block-brokers/src/trustless-gateway/utils.ts
+++ b/packages/block-brokers/src/trustless-gateway/utils.ts
@@ -20,6 +20,12 @@ export function filterNonHTTPMultiaddrs (multiaddrs: Multiaddr[], allowInsecure:
 
       return isPrivateIp(ma.toOptions().host) === false
     }
+    
+    // When allowInsecure is false and allowLocal is true, allow multiaddrs with "127.0.0.1", "localhost", or any subdomain ending with ".localhost"
+    if (!allowInsecure && allowLocal) {
+      if (ma.toOptions().host === '127.0.0.1' || ma.toOptions().host === 'localhost' || ma.toOptions().host.endsWith('.localhost'))
+        return true
+    }
 
     return false
   })
--- a/packages/block-brokers/src/trustless-gateway/utils.ts
+++ b/packages/block-brokers/src/trustless-gateway/utils.ts
@@ -20,11 +20,12 @@ export function filterNonHTTPMultiaddrs (multiaddrs: Multiaddr[], allowInsecure:
 
       return isPrivateIp(ma.toOptions().host) === false
     }
-    
+
     // When allowInsecure is false and allowLocal is true, allow multiaddrs with "127.0.0.1", "localhost", or any subdomain ending with ".localhost"
     if (!allowInsecure && allowLocal) {
-      if (ma.toOptions().host === '127.0.0.1' || ma.toOptions().host === 'localhost' || ma.toOptions().host.endsWith('.localhost'))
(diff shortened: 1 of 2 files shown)


packages/block-brokers/src/trustless-gateway/utils.ts
```diff
@@ -21,6 +21,13 @@ export function filterNonHTTPMultiaddrs (multiaddrs: Multiaddr[], allowInsecure:
       return isPrivateIp(ma.toOptions().host) === false
     }
 
+    // When allowInsecure is false and allowLocal is true, allow multiaddrs with "127.0.0.1", "localhost", or any subdomain ending with ".localhost"
+    if (!allowInsecure && allowLocal) {
+      if (ma.toOptions().host === '127.0.0.1' || ma.toOptions().host === 'localhost' || ma.toOptions().host.endsWith('.localhost')) {
```
SgtPooki, line 26: this looks okay, let's change it to

```suggestion
      const { host } = ma.toOptions()
      if (host === '127.0.0.1' || host === 'localhost' || host.endsWith('.localhost')) {
```

SgtPooki (changes requested):
lgtm, one suggested improvement. thanks a ton for throwing this together :)

SgtPooki (approved):

PR #584, opened 2024-08-21
fix: test for subscribers to ipns pubsub topic

lidel:
## Description

IPNS interop tests for helia→pubsub→kubo did not execute the same tests as kubo→pubsub→helia.

Namely, publishing to IPNS was executed with assumption it will fail, as a convoluted way of confirming the kubo is not subscribed to the topic, and also kubo connectivity state is somehow taken on belief, rather than being programmatically verified:

```js
// first publish should fail because kubo isn't subscribed to key update channel
await expect(name.publish(peerId, cid)).to.eventually.be.rejected()
  .with.property('message', 'PublishError.NoPeersSubscribedToTopic')

// should fail to resolve the first time as kubo was not subscribed to the pubsub channel
await expect(last(kubo.api.name.resolve(peerId, {
  timeout: 100
}))).to.eventually.be.undefined()
```

Good news is that there are native APIs for inspecting subsub topic subscriptions, and this PR refactors test to use them and have tests in both directions do more-or-less the same thing with the same asserts.

This should make interop less brittle.

## Notes & open questions

This  PR aims to fix problem identified in https://github.com/ipfs/kubo/pull/10488#discussion_r1725736598.

## Change checklist

### DIFF
--- a/packages/interop/src/ipns-pubsub.spec.ts
+++ b/packages/interop/src/ipns-pubsub.spec.ts
@@ -78,31 +78,39 @@ keyTypes.filter(keyType => keyType !== 'RSA').forEach(keyType => {
         throw new Error('No public key present')
       }
 
-      // first publish should fail because kubo isn't subscribed to key update channel
-      await expect(name.publish(peerId, cid)).to.eventually.be.rejected()
-        .with.property('message', 'PublishError.NoPeersSubscribedToTopic')
-
-      // should fail to resolve the first time as kubo was not subscribed to the pubsub channel
+      // first call to pubsub resolver will fail but we should trigger subscribing pubsub for updates
       await expect(last(kubo.api.name.resolve(peerId, {
         timeout: 100
       }))).to.eventually.be.undefined()
 
-      // magic pubsub subscription name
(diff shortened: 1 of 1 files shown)


lidel:
I did not mean to close this, bad Github.

SgtPooki (approved):
The code looks fine to me. Seems to just be validating that Kubo is actually subscribed like you mention in the PR description.

However, I'm confused why trying to resolve the IPNS name causes Kubo to subscribe to it. Is this a Kubo specific behavior or something in a spec somewhere?

lidel:
"IPNS over pubsub" is an example of alternative way of resolving IPNS (than Amino DHT).
There is relevant spec at https://specs.ipfs.tech/ipns/ipns-pubsub-router/

Modified test's Kubo runs with ["IPNS over PubSub" experiment](https://github.com/ipfs/kubo/blob/master/docs/experimental-features.md#ipns-pubsub) enabled:
https://github.com/ipfs/helia/blob/ab6b385787075ad9932f24362293b3bb82ff1d96/packages/interop/src/fixtures/create-kubo.ts#L31

This means once we resolve IPNS name, we keep subscription to a topic specific to the name, for getting updates faster than over DHT.

SgtPooki:
@achingbrain are you good with this change?

also, since this is a published "testing" package, I think it should be "fix: " in PR title

achingbrain (approved):

Issue #585, opened 2024-08-30
Helia incompatible with latest js-stores datastores


leblowl:
Due to https://github.com/ipfs/js-stores/pull/315, it appears that any code like this is currently incompatible: https://github.com/ipfs/helia/blob/ac4bdb8a73cab23500221340830969552a1d8db6/packages/utils/src/pins.ts#L157

For now, if you are using Helia 4.2.5 and Level (for example), it looks like datastore-level < 11.0.0 is required

SgtPooki:
@leblowl FYI: libp2p2.0 is coming soon and will also be changing error codes from `.code` to `.name` so we will likely not have this change out in a new Helia release until we can bring them both in

achingbrain:
The `.code` property was re-added to the js-stores implementations by https://github.com/ipfs/js-stores/pull/320 so this issue should be fixed now.  Sorry for the disruption.

The next Helia release will switch to using the `.name` property anyway but things are backwards compatible now.

achingbrain:
I'm going to close this issue, please re-open it if you're still having problems.

PR #588, opened 2024-08-31
fix: bitswap check err.name in peer-want-lists/ledger

dozyio:
## Description

Was hitting a weird edge case where `err.code` is not set. Adding an additional check for `err.name` fixes the issue.

I was only hitting the edge case with 3+ nodes using MemoryDatastore, 2 nodes seemed fine.

## Change checklist

### DIFF
--- a/packages/bitswap/src/peer-want-lists/ledger.ts
+++ b/packages/bitswap/src/peer-want-lists/ledger.ts
@@ -125,7 +125,10 @@ export class Ledger {
           })
         }
       } catch (err: any) {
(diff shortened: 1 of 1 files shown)


dozyio:
Not sure if you want to close this PR - didn't see https://github.com/ipfs/helia/issues/585 when I stated debugging

achingbrain:
Thanks for opening this - it's been fixed upstream by https://github.com/ipfs/js-stores/pull/320 so the breakage should be over now.

PR #590, opened 2024-09-06
chore: update benchmarks

achingbrain:
The bugs stopping WebRTC and WebTransport from being testable in browsers have been fixed and shipped so they can be added to the suite.

## Change checklist

### DIFF
--- a/benchmarks/transports/src/runner/helia/recipient.ts
+++ b/benchmarks/transports/src/runner/helia/recipient.ts
@@ -12,9 +12,9 @@ const cid = CID.parse(`${process.env.HELIA_CID}`)
 const mas = `${process.env.HELIA_MULTIADDRS}`.split(',').map(str => multiaddr(str))
 const signal = AbortSignal.timeout(parseInt(process.env.HELIA_TIMEOUT ?? '60000'))
 
-const helia = await getHelia()
-
 try {
+  const helia = await getHelia()
+
   await helia.libp2p.dial(mas, {
     signal
   })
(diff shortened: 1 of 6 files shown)


benchmarks/transports/src/tests.ts
```diff
@@ -99,6 +96,10 @@ const tcpImpls: Record<string, Impl> = {
 function addTests (name: string, impls: Record<string, Impl>, tests: Test[], relay: Multiaddr): void {
   for (const [implAName, implA] of Object.entries(impls)) {
     for (const [implBName, implB] of Object.entries(impls)) {
+      if (implAName !== implBName) {
+        continue
+      }
```
SgtPooki, line 101: this block will ensure we're only testing an implementation against itself. Is this supposed to be `===`?

SgtPooki (changes requested):

benchmarks/transports/src/tests.ts
achingbrain: Ah, no - left that in by accident

benchmarks/transports/src/tests.ts
```diff
@@ -99,6 +96,10 @@ const tcpImpls: Record<string, Impl> = {
...
+
```
achingbrain, line 102:
```suggestion
```

benchmarks/transports/src/runner/helia/recipient.ts
```diff
@@ -27,7 +27,8 @@ try {
   }))
 
   console.info(`TEST-OUTPUT:${Date.now() - start}`)
-} catch {
+} catch (err: any) {
+  console.error(err)
```
achingbrain, line 31:
```suggestion
} catch {
```

Issue #377, opened 2024-01-11
@helia/* modules should throw readable errors if CIDs with mismatched codecs are passed


achingbrain:
E.g. a CID suitable for use with `@helia/json` has the `.code` value `0x200` - if a CID without this codec is passed to `@helia/json`.get it should error saying something like:

> The passed CID had an incorrect codec, it may correspond to a non-JSON block

PR #643, opened 2024-10-03
fix: @helia/modules validate CID codec

tabcat:
## Title

fix: @helia/* modules validate CID codec

## Description

Closes: #377

@helia/* modules related to importing structured data, e.g. @helia/json; @helia/strings, check CID that is handed to `.get` method.
If the CID code does not match the codec of the importer then it rejects with a TypeError.

[x] - @helia/json
[ ] - @helia/strings
[ ] - @helia/dag-json
[ ] - @helia/dag-cbor
[ ] - @helia/unixfs

## Change checklist


packages/json/src/index.ts
```diff
@@ -115,6 +115,10 @@ class DefaultJSON implements JSON {
   }
 
   async get <T> (cid: CID, options: Partial<GetOptions> = {}): Promise<T> {
+    if (cid.code !== jsonCodec.code) {
+      throw new TypeError('The passed CID had an incorrect codec, it may correspond to a non-JSON block')
+    }
+
     const buf = await this.components.blockstore.get(cid, options)
```
tabcat, line 122: Checks that the codec codes match inside of the JSON.get implementation.

packages/json/src/index.ts
```diff
@@ -115,6 +115,10 @@ class DefaultJSON implements JSON {
   }
 
   async get <T> (cid: CID, options: Partial<GetOptions> = {}): Promise<T> {
```
tabcat, line 117: Random question, was looking at the GetOptions type. It contains `codec: BlockCodec` property which is passed to the `blockstore.get` call. Is this ever used?

tabcat:
Before I finish the other modules, does this approach look ok?

SgtPooki:
> Before I finish the other modules, does this approach look ok?

This looks good to me. It should be pretty straightforward

tabcat:
The `@helia/strings`.get method accepts a codec as an option. Either the option is removed (specifying a codec like raw, the current default) or validation cannot happen on the CID passed to get.

tabcat:
Actually I can just check the CID.code is raw if no codec option is passed to get.

tabcat:
Removing `@helia/unixfs` from todo list for this PR as it's a bit different and more involved than the other importers.

tabcat:
Depends on https://github.com/ipfs/helia/pull/668

I'll resolve the conflicts on that PR

PR #651, opened 2024-10-15
feat: enable filtering in delegated routing client

2color:
## Description

- **deps: upgrade delegated routing client**
- **feat: leverage filtering for delegated routing in libp2p constructor** 

## Change checklist

### DIFF
--- a/packages/helia/src/utils/libp2p-defaults.browser.ts
+++ b/packages/helia/src/utils/libp2p-defaults.browser.ts
@@ -64,7 +64,10 @@ export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOpti
     services: {
       autoNAT: autoNAT(),
       dcutr: dcutr(),
-      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev'),
+      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev', {
+        filterAddrs: ['unknown', 'transport-bitswap', 'transport-ipfs-gateway-http'],
+        filterProtocols: ['https', 'webtransport', 'webrtc', 'webrtc-direct', 'wss']
+      }),
       dht: kadDHT({
         clientMode: true,
         validators: {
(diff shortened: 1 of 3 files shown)


packages/helia/src/utils/libp2p-defaults.browser.ts
```diff
@@ -64,7 +64,10 @@ export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOpti
     services: {
       autoNAT: autoNAT(),
       dcutr: dcutr(),
-      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev'),
+      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev', {
+        filterProtocols: ['unknown', 'transport-bitswap', 'transport-ipfs-gateway-http'],
```
SgtPooki, line 68: what does the `unknown` filter protocol give us here? will that increase unreachable peers?

packages/helia/src/utils/libp2p-defaults.ts
```diff
@@ -73,7 +73,10 @@ export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOpti
     services: {
       autoNAT: autoNAT(),
       dcutr: dcutr(),
-      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev'),
+      delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev', {
+        filterProtocols: ['unknown', 'transport-bitswap', 'transport-ipfs-gateway-http'],
```
SgtPooki, line 77: ditto for `unknown`

packages/helia/src/utils/libp2p-defaults.browser.ts
2color: See https://github.com/ipfs/specs/pull/484/files#diff-2f9905a73a1926c9149e7bab39df73763be227e2d6f54fd16b708ee605369905R69

TL;DR: it's to include DHT providers, whose protocol isn't explicitly `bitswap` (even though in almost all cases they support bitswap)

packages/helia/src/utils/libp2p-defaults.ts
```diff
@@ -73,7 +73,10 @@ export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOpti
...
+        filterAddrs: ['https', 'tcp', 'webrtc', 'webrtc-direct', 'wss']
```
SgtPooki, line 78: we probably want to include `ws` & `http` here as well?

achingbrain: I can see that would be useful in test environments but probably not by default?

achingbrain (approved):

PR #652, opened 2024-10-15
fix: respect routers config in helia factory

2color:
## What's in this PR
- take the passed routers config in createHelia into acount

This PR allows overriding the routers configuration when customising Helia. Useful when you want to exclusively fetch directly from peers without relying on recursive gateways.

## Relevant context

In `@helia/http` we already do this:
https://github.com/ipfs/helia/blob/332580c4dd1646381bfeedd7741e52bc3a364301/packages/http/src/index.ts#L79-L82


SgtPooki (approved):
lgtm!

PR #654, opened 2024-10-16
feat: enable customising delegated http router

2color:
## Description

Allow passing delegated routing client options, e.g. IPIP-484 filters, to the delegated HTTP router.

## Notes & open questions

- Should we also set default filters as proposed in #651 to the delegated HTTP router? If so, should we follow the same defaults as #651 (with different defaults for browser vs. node.js)?

## Change checklist


packages/routers/src/delegated-http-routing.ts
```diff
@@ -94,6 +94,6 @@ class DelegatedHTTPRouter implements Routing {
 /**
  * Creates a Helia Router that connects to an endpoint that supports the [Delegated Routing V1 HTTP API](https://specs.ipfs.tech/routing/http-routing-v1/) spec.
  */
-export function delegatedHTTPRouting (url: string | URL): Routing {
-  return new DelegatedHTTPRouter(new URL(url))
+export function delegatedHTTPRouting (url: string | URL, init: DelegatedRoutingV1HttpApiClientInit = {}): Routing {
+  return new DelegatedHTTPRouter(new URL(url), init)
 }
```
SgtPooki, line 99: we probably also need to update the `createHelia` functions to support passing these through right?

SgtPooki (approved):

packages/routers/src/delegated-http-routing.ts
2color: For the normal `helia`, i .e. HeliaP2P, we don't even instantitate the delegatedHTTPRouter, so not relevant:
https://github.com/ipfs/helia/blob/e5c03bdb7a1f64f78594dade0571eef4abcb098c/packages/helia/src/index.ts#L115-L130

For `@helia/http`, we do use the `delegatedHTTPRouter`: https://github.com/ipfs/helia/blob/f904041bdd2643fabd204485825d4761fc44a83f/packages/http/src/index.ts/#L79-L82

But if you want to cutsomise the delegatedHTTPRouter options, you can just instantiate it yourself and pass it in via `init.router` (a bit like how we do it in [the example](https://github.com/ipfs-examples/helia-examples/blob/4d8192e4b3d95a736e9e88ce97568ac09bff830f/examples/helia-browser-verified-fetch/src/utils.ts#L51-L52))

SgtPooki:
We need to get the CI passing here so we can unblock changes in https://github.com/ipfs/service-worker-gateway/pull/391

packages/routers/src/delegated-http-routing.ts
```diff
@@ -6,7 +6,8 @@ import map from 'it-map'
 import { CID } from 'multiformats/cid'
 import { equals as uint8ArrayEquals } from 'uint8arrays/equals'
 import { fromString as uint8ArrayFromString } from 'uint8arrays/from-string'
-import type { DelegatedRoutingV1HttpApiClient } from '@helia/delegated-routing-v1-http-api-client'
+import { delegatedHTTPRoutingDefaults } from './utils/delegated-http-routing-defaults'
```
achingbrain, line 9:
```suggestion
import { delegatedHTTPRoutingDefaults } from './utils/delegated-http-routing-defaults.js'
```

achingbrain (approved):

PR #668, opened 2024-10-28
refactor(strings)!: remove codec from AddOptions

tabcat:
## Title

refactor(strings)!: remove codec from AddOptions

## Description

This changes @helia/strings to act more like the other helia importers.
It simplifies the `@helia/strings` importer to only encode strings as 'utf8' and apply the raw codec to output CIDs.
If users need to import data to helia using a different codec, one of the other helia importers (json, dag-json, dag-cbor) can be used.
For special use cases a user can handle the encoding/decoding, hashing, and then importing themselves. `multiformats`' Block.encode/decode is helpful for this. Happy to add an example with this pull.

## Notes & open questions

This would be a prerequisite to #643

BREAKING CHANGE:

Removes `codec` from AddOptions for `@helia/strings.add`. Anyone relying
on this option currently would have to encode, hash, and import the data
using helia's blockstore (unless implemented by another helia importer
e.g. `@helia/json` or `@helia/dag-cbor`) in order to receive the same
CID as before.

## Change checklist


SgtPooki:
@tabcat do you still have time to complete this and #643 ?

tabcat:
Does this need documentation changes to be merged? I added description to the initial commit that I thought may make it to the Changelog but I'm not too familiar with the conditions for that to work.

achingbrain (approved):

PR #670, opened 2024-10-30
fix: add tls to default delegated routing filters

2color:
## Description

- **fix: add tls to all default filters**
- **chore: avoid duplication of filters**

## Notes & open questions

Should we just move these defaults to the [Delegated Routing client](https://github.com/ipfs/helia-delegated-routing-v1-http-api/blob/main/packages/client/src/client.ts#L26)? 

## Change checklist

### DIFF
--- a/packages/helia/src/utils/libp2p-defaults.browser.ts
+++ b/packages/helia/src/utils/libp2p-defaults.browser.ts
@@ -65,7 +65,7 @@ export function libp2pDefaults (options: Libp2pDefaultsOptions = {}): Libp2pOpti
       dcutr: dcutr(),
       delegatedRouting: () => createDelegatedRoutingV1HttpApiClient('https://delegated-ipfs.dev', {
         filterProtocols: ['unknown', 'transport-bitswap', 'transport-ipfs-gateway-http'],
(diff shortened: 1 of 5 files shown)


SgtPooki (approved):
Youll need to add the browser default to the package.json to ensure its picked up

2color:
> Youll need to add the browser default to the package.json to ensure its picked up

We're already doing that in the `@helia/routers`  pacakge: 
https://github.com/ipfs/helia/blob/3f8558a87735c74a2b4907c9b7de83b5fa6dd5c6/packages/routers/package.json#L76

Is it necessary also here?

achingbrain (approved):

achingbrain:
> Should we just move these defaults to the [Delegated Routing client](https://github.com/ipfs/helia-delegated-routing-v1-http-api/blob/main/packages/client/src/client.ts?rgh-link-date=2024-10-30T13%3A18%3A12Z#L26)?

Probably not?  The defaults here are tied to the libp2p configuration (as in, no point adding the TLS filter if you've not got `@libp2p/tls` in your setup) which would be hard to predict in the delegated routing client, probably best to just let people configure it for their needs.

PR #674, opened 2024-11-07
fix: remove webtransport from default transports

2color:
## Description

WebTransport is still working reliably across different browsers. Moreover, the go-libp2p implementation needs to be updated to meet newer revisions of the spec. 

As we discussed during Helia WG, the most sensible thing for now is to remove WebTransport from the default transports.

Related to https://github.com/libp2p/js-libp2p/issues/2572

## Change checklist

### DIFF
--- a/benchmarks/transports/package.json
+++ b/benchmarks/transports/package.json
@@ -25,7 +25,6 @@
     "@libp2p/tcp": "^10.0.0",
     "@libp2p/webrtc": "^5.0.0",
     "@libp2p/websockets": "^9.0.0",
-    "@libp2p/webtransport": "^5.0.0",
     "@multiformats/multiaddr": "^12.2.1",
     "aegir": "^45.0.1",
     "blockstore-fs": "^2.0.1",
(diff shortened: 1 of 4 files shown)


benchmarks/transports/package.json
```diff
@@ -25,7 +25,6 @@
     "@libp2p/tcp": "^10.0.0",
     "@libp2p/webrtc": "^5.0.0",
     "@libp2p/websockets": "^9.0.0",
-    "@libp2p/webtransport": "^5.0.0",
```
SgtPooki, line 28: idk that we should remove it from benchmarks?

SgtPooki (changes requested):
I would say the only change we really want are the ones to `packages/helia`

SgtPooki (approved):

2color:
@achingbrain Any thoughts on this?

achingbrain:
This makes me a bit 😭 but I don't how else we can give users a good experience otherwise right now.

PR #720, opened 2025-01-17
chore: create FUNDING.json

cewood:
Add FUNDING.json indicating that this repo is owned/maintained by Shipyard.

### DIFF
--- a/FUNDING.json
+++ b/FUNDING.json
@@ -1,0 +1,7 @@
+{
+  "drips": {
+    "filecoin": {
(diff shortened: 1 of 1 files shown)


FUNDING.json
```diff
@@ -0,0 +1,7 @@
+{
+  "drips": {
+    "filecoin": {
```
2color, line 3: Should we also add an Ethereum one?

cewood: I don't think so, let's get this merged for now so we can claim the DRIPs we've been granted. And we can add other ones as needed later, if that happens.

achingbrain (approved):

Issue #721, opened 2025-01-21
IPNS cannot resolve DNSLink with ipns public keys


RangerMauve:
In Helia 5.2.0

Reproducible example:

```javascript
const helia = await createHelia({
    libp2p: {
        start: false
    }
})

const ipns = createIPNS(helia)

const dnsLink = await ipns.resolveDNSLink('hypha.coop')

console.log({dnsLink})
```

The issue is that the returned record is in the form of `/ipns/{public key}` and the code for [parsing](https://github.com/ipfs/helia/blob/main/packages/ipns/src/dnslink.ts#L69) the "peerID" isn't passing in a MultibaseDecoder [to the parseIdFromString](https://github.com/libp2p/js-libp2p/blob/main/packages/peer-id/src/index.ts#L33) API.

Here's what our TXT record looks like: `dnslink=/ipns/k51qzi5uqu5dlmj8c05a6xk0g0n09x2m3a76lq3ahemgwtsr8s0r4lax9f5yvp/`

IMO the 'peerIdFromString' function is not appropriate here and a CID parser or MultibaseDecoder would be more appropriate. But just passing a decoder should be sufficient.

RangerMauve:
Just tested and resolving works fine in `helia@4.2` and `@helia/ipns@7`

2color:
I believe we had a similar problem in Verified Fetch, which was caused by a breaking change to `peerIdFromString` whereby it no longer tried to parse it as a CID. 

We solved it by creating a wrapper:
https://github.com/ipfs/helia-verified-fetch/blob/main/packages/verified-fetch/src/utils/get-peer-id-from-string.ts#L5-L12

RangerMauve:
Thank you for the quick fix!

PR #722, opened 2025-01-22
fix: handle dnslink pointing to  CID with peer id (ipns name)

2color:
## What's in this PR
fixes https://github.com/ipfs/helia/issues/721

## Change checklist


achingbrain (changes requested):
Thanks - can you please add a test to ensure there are no future regressions?

2color:
Tests are in place

SgtPooki (approved):
lgtm. the test seems legit

PR #723, opened 2025-01-22
feat: allow passing custom ttl when publishing ipns

2color:
## What's in this PR

- **chore: export IPNS Record type**
- **feat: allow setting ttl in miliseconds**
- **test: add test for custom ttl**

## Change checklist

### DIFF
--- a/packages/ipns/test/publish.spec.ts
+++ b/packages/ipns/test/publish.spec.ts
@@ -48,15 +48,35 @@ describe('publish', () => {
     expect(ipnsEntry).to.have.property('ttl', 3_600_000_000_000n) // 1 hour
   })
 
-  it('should publish an IPNS record with a custom ttl params', async function () {
+  it('should publish an IPNS record with a custom lifetime params', async function () {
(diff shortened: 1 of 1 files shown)


packages/ipns/src/index.ts
```diff
@@ -429,6 +434,8 @@ export interface IPNS {
 
 export type { IPNSRouting } from './routing/index.js'
 
+export type { IPNSRecord } from 'ipns'
```
SgtPooki, line 437: I wonder if we want to export this or have folks import it from 'ipns' directly. If it's just a type, it shouldn't cause bundle-bloat issues.

SgtPooki (commented):
I'm good with this, just one thought

SgtPooki (approved):

packages/ipns/src/index.ts
2color: Yeah to be honest, this one was really needed and the only thing I ended up needing to import from `ipns` so I decided it's used enough to be reexported.

PR #732, opened 2025-02-01
feat: add support for custom headers in trustless gateway

pcfreak30:


SgtPooki (commented):
@pcfreak30 Thanks again for throwing this together.

I talked with @achingbrain about this. We will probably alter this to accept a function where users can alter the RequestInit object we pass to the `fetch` call on line 108 of `packages/block-brokers/src/trustless-gateway/trustless-gateway.ts`

This will allow users to modify headers or anything else they need to so we don't have to foresee the various changes users might need to make, nor keep modifying or maintaining those changes as needed for their special use-cases.

This will keep things flexible for various new use-cases while keeping the maintenance burden low

SgtPooki:
closing this in favor of #751. Thanks for the headstart @pcfreak30

PR #743, opened 2025-02-23
docs: add spell checker to ci

achingbrain:
Adds cspell to CI to catch typos in code and docs.

## Change checklist


packages/bitswap/src/want-list.ts
```diff
@@ -57,12 +57,12 @@ export interface WantListEntry {
   /**
    * Whether the remote should tell us if they have the block or not
    */
-  sendDontHave: boolean
+  sendDoNotHave: boolean
```
2color, line 60: Since we export this interface, does this qualify as a breaking change?

package.json
```diff
@@ -31,14 +31,15 @@
     "lint": "aegir run lint",
     "dep-check": "aegir run dep-check",
     "doc-check": "aegir run doc-check",
+    "spell-check": "aegir spell-check",
```
2color, line 34: Should we add a `check` script that runs all 4:
```
    - npm run lint
    - npm run dep-check
    - npm run doc-check
    - npm run spell-check
```

Specifically, this seems like it would be useful for external contributors, when checks fails to run, they can fix locally quickly, without familiarity with each of the scripts.

2color (approved):
Nice

Issue #744, opened 2025-02-24
Add ability to republish IPNS records without private key


2color:
## Background

The current `@helia/ipns` API assumes that you hold the private key for IPNS names.  There are some situations where you might want to republish a valid record that was created by another node, however we're missing the API for it.

## API Ideas

```ts
republishRecord(record: IPNSRecord, pubKey: PublicKey, options?: RepublishRecordOptions): Promise<void>
```

PR #745, opened 2025-02-24
feat: add republish signed ipns records

2color:
## What

Fixes https://github.com/ipfs/helia/issues/744

## Change checklist


packages/ipns/src/index.ts
```diff
@@ -430,6 +437,13 @@ export interface IPNS {
    * Periodically republish all IPNS records found in the datastore
    */
   republish(options?: RepublishOptions): void
+
+  /**
+   * Republish an existing IPNS record without the private key
+   *
+   * The public key is optional if the record has an embedded public key.
+   */
+  republishRecord(record: IPNSRecord, pubKey?: PublicKey, options?: RepublishRecordOptions): Promise<void>
```
SgtPooki, line 446: stick pubKey into `RepublishRecordOptions` object?

packages/ipns/src/index.ts
```diff
@@ -707,6 +721,33 @@ class DefaultIPNS implements IPNS {
 
     return unmarshalIPNSRecord(record)
   }
+
+  async republishRecord (record: IPNSRecord, pubKey?: PublicKey, options: RepublishRecordOptions = {}): Promise<void> {
```
SgtPooki, line 725: ditto

packages/ipns/test/republish.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+      {
+        routers: [customRouting]
+      }
+    )
+
+    testCid = CID.parse('QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn')
+    rsaKey = await generateKeyPair('RSA') // RSA will embed the public key in the record
+    ed25519Key = await generateKeyPair('Ed25519')
+    rsaRecord = await createIPNSRecord(rsaKey, testCid, 1n, 24 * 60 * 60 * 1000)
+    ed25519Record = await createIPNSRecord(ed25519Key, testCid, 1n, 24 * 60 * 60 * 1000)
```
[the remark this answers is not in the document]
SgtPooki: just a nit here, but testCid could probably be moved up into the describe block, and then the records and keys could be moved into individual tests

achingbrain (approved):
LGTM - just needs @SgtPooki's comments addressing and it's ready.

packages/ipns/src/index.ts
2color: I thought about this for a little bit. Since ed25519 account for most of the IPNS names in the network, and the public key isn't embedded into the IPNS record. 

Given that, I think it's better if it remains as an argument rather than an option.

2color: > Could we make the key required instead?

That makes sense. For embedded pub keys consumers should be able to easily pass those in. And it makes sense to design the external APIs around ed25519 which are most common.

> Since we aren't signing the record, do we really need a PublicKey or will a Multihash we can turn into the routing key do?

multihash should do. 

> Are there many situations where we want to republish an IPNS record and we don't expect to have the routing key separate to the record?

This came up in the context of republishing IPNS records in https://github.com/ipshipyard/ipns-inspector. Though we can address it by embedding in the record (for ui authored records) and in the filename (for looked up records).

2color: Done.

2color:
I've made some changes to address @achingbrain's feedback from https://github.com/ipfs/helia/pull/745#discussion_r1987265416, taking in a key first and the value second. I also added the ability to pass in a string

packages/ipns/src/index.ts
```diff
@@ -379,6 +380,13 @@ export interface RepublishOptions extends AbortOptions, ProgressOptions<Republis
   interval?: number
 }
 
+export interface RepublishRecordOptions extends AbortOptions, ProgressOptions<RepublishProgressEvents | IPNSRoutingEvents> {
+  /**
+   * Only publish to a local datastore (default: false)
```
achingbrain, line 385: Tells tsdoc what the default value is:

```suggestion
   * Only publish to a local datastore
   *
   * @default false
```

packages/ipns/src/index.ts
```diff
@@ -707,6 +722,58 @@ class DefaultIPNS implements IPNS {
...
+  async republishRecord (key: MultihashDigest<0x00 | 0x12> | string, record: IPNSRecord, options: RepublishRecordOptions = {}): Promise<void> {
+    let mh: MultihashDigest<0x00 | 0x12> | undefined
+    try {
+      mh = extractPublicKeyFromIPNSRecord(record)?.toMultihash() // embedded public key take precedence, if present
+      if (mh == null) {
+        // if no public key is embedded in the record, use the key that was passed in
+        if (typeof key === 'string') {
+          // Convert string key to MultihashDigest
+          try {
+            mh = this.#getPeerIdFromString(key).toMultihash()
```
achingbrain, line 748: Should we require string keys to be fully qualified?  E.g. `/ipns/...`

packages/ipns/src/index.ts
```diff
@@ -302,7 +303,7 @@ export type ResolveProgressEvents =
 export type RepublishProgressEvents =
   ProgressEvent<'ipns:republish:start', unknown> |
   ProgressEvent<'ipns:republish:success', IPNSRecord> |
-  ProgressEvent<'ipns:republish:error', { record: IPNSRecord, err: Error }>
+  ProgressEvent<'ipns:republish:error', { mh?: MultihashDigest<0x00 | 0x12>, record: IPNSRecord, err: Error }>
```
achingbrain, line 306: `mh` is a bit vague, maybe call it `key` to match the function args?

packages/ipns/test/republish.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        routers: [customRouting]
+      }
+    )
+  })
+
+  it('should throw an error when attempting to republish with an invalid key', async () => {
+    const ed25519Key = await generateKeyPair('Ed25519')
+    const otherEd25519Key = await generateKeyPair('Ed25519')
+    const ed25519Record = await createIPNSRecord(ed25519Key, testCid, 1n, 24 * 60 * 60 * 1000)
+    await expect(name.republishRecord(otherEd25519Key.publicKey.toMultihash(), ed25519Record)).to.be.rejected
```
achingbrain, line 49: When asserting rejection it's a good idea to assert which error was thrown:

```suggestion
    await expect(name.republishRecord(otherEd25519Key.publicKey.toMultihash(), ed25519Record)).to.be.rejected
      .with.property('name', 'TheExpectedError')
```

achingbrain (changes requested):
Looking good, final few nits

packages/ipns/src/index.ts
2color: mmmmm. According to the spec that's good practice but not required. I'd rather not, since the current APIs don't expect it, so this would be a bit contrived.

SgtPooki (approved):
👍 lgtm...

packages/ipns/src/index.ts
```diff
@@ -707,6 +747,58 @@ class DefaultIPNS implements IPNS {
 
     return unmarshalIPNSRecord(record)
   }
+
+  /**
+   * Convert a string to a PeerId
+   */
+  #getPeerIdFromString (peerIdString: string): PeerId {
```
2color, line 754: Once https://github.com/libp2p/js-libp2p/pull/3042 is released, we can just use `peerIdFromString`

achingbrain: It has shipped

packages/ipns/src/index.ts
achingbrain: We should probably strip any leading `/ipns/` from the key then?

2color: Will do in a follow up PR

PR #747, opened 2025-02-26
fix!: fix typo in bitswap

achingbrain:
Replaces `sendDontHave` with `sendDoNotHave` because `Dont` is not a word.

BREAKING CHANGE: The `sendDontHave` field of the `WantListEntry` is now `sendDoNotHave`

## Change checklist


2color (approved):

dozyio:
I think using `DontHave` is ok as that is what is defined in the spec

achingbrain:
Groan, fair enough.

I wish we had a spell checker for our specs.  Like how the libp2p transport interop test has us returning [pingRTTMilllis](https://github.com/libp2p/test-plans/tree/master/transport-interop#dialer) (3x Ls).

PR #748, opened 2025-02-26
feat: downloadOnly option for bitswap

dozyio:
## Description

Add download only option to bitswap

## Notes & open questions

Thinking this should be an option for mobile clients on metered connections and also for privacy reasons.

## Change checklist


dozyio:
Seems Go bitswap uses a PeerBlockRequestFilter https://github.com/ipfs/boxo/blob/main/bitswap/server/internal/decision/engine.go#L252 - might be better than downloadOnly as more general purpose - any thoughts?

PR #760, opened 2025-03-11
fix!: return simple stats or extended stats

### LINKED ISSUE #580: UnixFs Stat Command Returns Invalid Data
When using stat, it does not return the correct dagSize values for non-empty directories. It appears that [this line](https://github.com/ipfs/helia/blob/main/packages/unixfs/src/commands/stat.ts#L54) is using the byteLength of the UnixFS data in the directory block. This causes stat commands on non-empty directories to return an invalid value that does not match the Kubo value.

achingbrain:
Splits the unixfs/mfs stat command return types into two types - "regular" stats (these only return stats available on the root node of the DAG) and "extended" - these collect stats from the DAG, potentially going to the network to fetch missing blocks.

There are separate simple and extended types for files and dirs, but raw/leaf nodes only get one type since there are never any nodes linked to from them.

As a bonus we can now calculate the size of directories correctly, (as the combination of all child files/directories), assuming all blocks are present in the blockstore or fetchable from the network.

Fixes #580

BREAKING CHANGE: Fields that would involve DAG traversal have been removed from the output of `fs.stat` - pass the `extended` option to have them returned

## Change checklist


achingbrain:
Check out the changes to the returned types - https://github.com/ipfs/helia/pull/760/files#diff-026d52cd110beb421b24bb2d95961f27f3f3d083df76eac8302eb4eab783c680 - most of this PR is noise dealing with that.

achingbrain:
Removing the `mtime`/`mode` fields from the stat results might be a bit much since there are supposed to be default values if they aren't specified in the UnixFS metadata, will probably reinstate them which will make this PR a lot quieter.

packages/mfs/test/stat.spec.ts
```diff
@@ -23,52 +23,69 @@ describe('stat', () => {
...
-      blocks: 5,
-      localFileSize: 4194304n,
-      localDagSize: 4194563n
+    const updatedStats = await fs.stat(filePath, {
+      extended: true
     })
+
+    expect(updatedStats.unixfs?.fileSize()).to.equal(5242880n)
+    expect(updatedStats.blocks).to.equal(5n)
+    expect(updatedStats.dagSize).to.equal(4194563n)
```
SgtPooki, line 68: I was expecting dagSize to be the same here because `extended=true` but I guess there's no networking done in the test, so it doesn't retrieve that block?

packages/unixfs/src/commands/stat.ts
```diff
@@ -1,112 +1,134 @@
...
+    }
+
+    return createStats(result)
+  } else if (result.type === 'directory') {
+    if (options.extended === true) {
+      return createExtendedStats(result, blockstore, options)
+    }
+
+    return createStats(result)
   }
```
SgtPooki, line 53: I believe these two if branches can be merged. e.g. `if (type==='file' || type === 'directory')`

SgtPooki (approved):
the only thing confusing me here is that in the tests, we aren't validating that "network requests" will fill out `dagSize` more when `extended=true`. 

It would be good to have a test for that.. maybe have a wrapped blockstore that calls to a "network" blockstore that we load the bytes into?

packages/mfs/test/stat.spec.ts
achingbrain: That's correct, yes. The blockstore is a MemoryBlockstore that isn't wrapped in a NetworkedBlockstore.

2color:
> the only thing confusing me here is that in the tests, we aren't validating that "network requests" will fill out dagSize more when extended=true.

I was confused by the same thing.

If you have a NetworkedBlockstore, will the `await blockstore.has(cid, options))` trigger a chain leading to a find providers calls and `const block = await blockstore.get(cid, options)` call to fetch the block?  
https://github.com/ipfs/helia/blob/6a9f2976711b25ffd69414c0b7ebc1ebdd865f90/packages/unixfs/src/commands/stat.ts#L108-L118

packages/unixfs/src/commands/stat.ts
```diff
@@ -1,112 +1,124 @@
...
-    localDagSize = BigInt(result.node.byteLength)
-    blocks = 1
+function createStats (entry: UnixFSFile | UnixFSDirectory): FileStats | DirectoryStats {
+  return {
+    type: entry.type,
+    cid: entry.cid,
+    unixfs: entry.unixfs,
+    mode: entry.unixfs.mode ?? (entry.unixfs.isDirectory() ? DEFAULT_DIR_MODE : DEFAULT_FILE_MODE),
+    mtime: entry.unixfs.mtime,
+    size: entry.unixfs.fileSize()
```
2color, line 59: What do we expect `fileSize()` to return for directories?

packages/unixfs/src/index.ts
```diff
@@ -265,49 +276,88 @@ export interface UnixFSStats {
...
-  blocks: number
+  dagSize: bigint
 
   /**
-   * The type of file
+   * How much of the file or directory is in the local block store. If this is a
+   * directory it will include the size of all child files and directories.
    */
-  type: 'file' | 'directory' | 'raw'
+  localSize: bigint
```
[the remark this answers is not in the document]
achingbrain: Yes, unless the `offline` option was `true`, in which case it won't pull down and missing blocks so `localSize` will only be what's in the local blockstore.

achingbrain:
> If you have a NetworkedBlockstore, will the await blockstore.has(cid, options)) trigger a chain leading to a find providers calls and const block = await blockstore.get(cid, options) call to fetch the block?

No, actually this was a bug.  I've updated the code and added an interop test that tests pulling the missing block from a network peer.

packages/unixfs/src/index.ts
```diff
@@ -265,49 +276,88 @@ export interface UnixFSStats {
...
-   * The size of the DAG that holds the file in bytes
+   * UnixFS metadata about this file or directory
    */
-  dagSize: bigint
+  unixfs?: IPFSUnixFS
 
   /**
-   * How much of the file is in the local block store
+   * The size of the filesystem entry. By default this is accurate for 'file'
+   * and 'raw' entries, for directories please stat with the `extended` option.
```
2color, line 298:
```suggestion
   * and 'raw' entries. For directories, please stat with the `extended` option, as this will return 0.
```

2color:
To better understand and doucment the nuances around stats and extended stats, I created this table:


## stat options and result

| input CID type | offline | extended | fetched block(s) | DAG traversal | size       | localSize  | dagSize                 | blocks |
| -------------- | ------- | -------- | ---------------- | ------------- | ---------- | ---------- | ----------------------- | ------ |
| file           | ❌   | ❌    | ✅             | ❌         | file size  | n/a        | n/a                     | n/a    |
| directory      | ❌   | ❌    | ✅             | ✅          | 0          | n/a        | n/a                     | n/a    |
| raw            | ❌   | ❌    | ✅             | n/a           | bytelength | n/a        | n/a                     | n/a    |
| file           | ✅    | ❌    | ❌            | ❌         | filesize   | n/a        | n/a                     | n/a    |
| directory      | ✅    | ❌    | ❌            | ❌         | 0          | n/a        | n/a                     | n/a    |
| raw            | ✅    | ❌    | ❌            | n/a           | bytelength | n/a        | n/a                     | n/a    |
| file           | ❌   | ✅     | ✅             | ✅          | file size  | file size  | file size + pb overhead | blocks |
| directory      | ❌   | ✅     | ✅             | ✅          | dir size   | 0          | dir size + pb overhead  | blocks |
| raw            | ❌   | ✅     | ✅             | n/a           | bytelength | bytelength | bytelength              | 1      |
| file           | ✅    | ✅     | ❌            | ✅          | file size  | file size  | file size + pb overhead | blocks |
| directory      | ✅    | ✅     | ❌            | ✅          | dir size   | 0          | dir size + pb overhead  | blocks |
| raw            | ✅    | ✅     | ❌            | n/a           | bytelength | bytelength | bytelength              | 1      |

## Notes
 - With `{ extended: true, offline: true }`, the difference between `size` and `localSize` is the delta between local blocks and total blocks of the dag.
- For files with normal stats, size will be calculated from the unixfs metadata.
- For directories with normal stats, size will be 0n.
- For directories with extended stats, `size` and `localSize` will be the same. If some blocks are missing, both will be inaccurate.

## 
I'll drop it here for now, and can incorporate it into the READMEs in a follow up to this. 

Would be great if you can verify this.

achingbrain:
> Would be great if you can verify this.

Looks fine.  I've made a change so that when statting a directory with `extended: true`, `size` is the recursive sum of the `size` property of all child files (where the file's DAG root is in the blockstore) and not 0.

This is as close to the "true" or expected directory size as we can get, I think.

For the same stat result, `localSize` is the recursive sum of the `localSize` property of all child files.

PR #761, opened 2025-03-11
docs: update @helia/car export method example

SgtPooki:
## Title

docs: update @helia/car export method example

## Description

Fixes an issue with the example with invalid code.

* removed `await` from `c.export` because that results in an error in node.js v22.9.0:

```
Warning: Detected unsettled top-level await at file:///Users/sgtpooki/code/work/protocol.ai/ipfs/helia/packages/car/test.js:29
await c.export(cid, writer)
^
```

* removed pEvent listening on wrong (`end`) event, which doesn't exist in LTS: https://nodejs.org/docs/latest-v22.x/api/stream.html

## Notes & open questions

Tested with code locally:

```js
import fs from 'node:fs'
import path from 'node:path'
import { cwd } from 'node:process'
import { Readable } from 'stream'
import { car } from '@helia/car'
import { createHeliaHTTP } from '@helia/http'
import { createVerifiedFetch } from '@helia/verified-fetch'
import { CarWriter } from '@ipld/car'
import { FsBlockstore } from 'blockstore-fs'
import { FsDatastore } from 'datastore-fs'

const currentPath = path.join(cwd(), 'packages', 'car')
const helia = await createHeliaHTTP({
  blockstore: new FsBlockstore(path.join(currentPath, 'blockstore')),
  datastore: new FsDatastore(path.join(currentPath, 'datastore'))
})

const verifiedFetch = await createVerifiedFetch(helia)

await verifiedFetch('https://ipfs.io/ipfs/QmSnuWmxptJZdLJpKRarxBMS2Ju2oANVrgbr2xWbie9b2D/_Metadata.json?format=raw')

const cid = CID.parse('QmWXShtJXt6Mw3FH7hVCQvR56xPcaEtSj4YFSGjp2QxA4v')

const c = car(helia)

const { writer, out } = CarWriter.create(cid)

c.export(cid, writer)

const output = fs.createWriteStream('example.car')

await new Promise((resolve, reject) => {
  const stream = Readable.from(out).pipe(output)
  stream.on('finish', () => {
    resolve()
  })
  stream.on('error', (err) => {
    reject(err)
  })
})

// check that example.car exists
const exists = fs.existsSync('example.car')
if (!exists) {
  throw new Error('example.car does not exist')
}
```

### Some questions:

1. Do we need c.export to be a promise?
2. Should we update the example to listen to `out.on('finish'` instead?

## Change checklist

### DIFF
--- a/packages/car/src/index.ts
+++ b/packages/car/src/index.ts
@@ -120,31 +120,28 @@ export interface Car {
    *
    * ```typescript
    * import fs from 'node:fs'
-   * import { Readable } from 'stream'
+   * import { Readable } from 'node:stream'
    * import { car } from '@helia/car'
    * import { CarWriter } from '@ipld/car'
    * import { createHelia } from 'helia'
    * import { CID } from 'multiformats/cid'
+   * import { pEvent } from 'p-event'
    *
    * const helia = await createHelia()
    * const cid = CID.parse('QmFoo...')
    *
    * const c = car(helia)
    * const { writer, out } = CarWriter.create(cid)
-   * c.export(cid, writer).catch(console.error)
    * const output = fs.createWriteStream('example.car')
-   * await new Promise((resolve, reject) => {
-   *   const stream = Readable.from(out).pipe(output)
-   *   stream.on('finish', () => {
-   *     resolve()
-   *   })
-   *   stream.on('error', (err) => {
-   *     reject(err)
-   *   })
-   * })
+   * const stream = Readable.from(out).pipe(output)
+   *
+   * await Promise.all([
+   *   c.export(cid, writer),
+   *   pEvent(stream, 'close')
+   * ])
+   * ```
    *
    * @deprecated Use `stream` instead. In a future release `stream` will be renamed `export`.
-   * ```
    */
   export(root: CID | CID[], writer: Pick<CarWriter, 'put' | 'close'>, options?: ExportCarOptions): Promise<void>
 
--- a/packages/car/src/index.ts
+++ b/packages/car/src/index.ts
@@ -121,22 +121,27 @@ export interface Car {
    * ```typescript
    * import fs from 'node:fs'
    * import { Readable } from 'stream'
+   * import { car } from '@helia/car'
+   * import { CarWriter } from '@ipld/car'
    * import { createHelia } from 'helia'
-   * import { car } from '@helia/car
    * import { CID } from 'multiformats/cid'
-   * import pEvent from 'p-event'
    *
    * const helia = await createHelia()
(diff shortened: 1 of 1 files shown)


SgtPooki (commented):
self review

achingbrain:
Maybe we should deprecate* that method as it's a bit awkward.

We added the `stream` method in #444 which feels a bit more natural:

```js
const c = car(helia)

// where the car bytes will end up
const output = fs.createWriteStream('example.car')

// stream car bytes to file
Readable.from(c.stream(cid)).pipe(output)

// wait for output stream to either error or finish writing
await pEvent(output, 'close')
```


----

\* = Probably remove `export` and rename `stream` to `export` in a breaking release to preserve the `import`/`export` symmetry.

achingbrain:
```
Warning: Detected unsettled top-level await at file:///Users/sgtpooki/code/work/protocol.ai/ipfs/helia/packages/car/test.js:29
await c.export(cid, writer)
^
```

So this is an interesting error message.  It means all async work has finished, so the event loop will exit but we're still waiting for a promise to resolve.

I think this might be because with `await c.export(cid, writer)` it means we wait for the export to finish before something has started to read from the output stream.

That's probably enough to upgrade the `.export` method to "considered harmful".

Background reading: https://stevenwmerrill.com/blog/2025-01-27-i-promise-you-shouldn-t-fear-rejection/

packages/car/src/index.ts
```diff
@@ -121,22 +121,27 @@ export interface Car {
...
-   * await eventPromise
+   * await new Promise((resolve, reject) => {
+   *   const stream = Readable.from(out).pipe(output)
+   *   stream.on('finish', () => {
+   *     resolve()
+   *   })
+   *   stream.on('error', (err) => {
+   *     reject(err)
+   *   })
+   * })
```
achingbrain, line 144:
```suggestion
   * })
   *
   * @deprecated Use `stream` instead. In a future release `stream` will be renamed `export`.
```

packages/car/src/index.ts
```diff
@@ -121,22 +121,27 @@ export interface Car {
...
-   * import pEvent from 'p-event'
    *
    * const helia = await createHelia()
    * const cid = CID.parse('QmFoo...')
    *
    * const c = car(helia)
-   *
-   * const byteStream = await c.export(cid)
+   * const { writer, out } = CarWriter.create(cid)
+   * c.export(cid, writer)
```
achingbrain, line 134: Prevent unhandled rejections:

```suggestion
   * c.export(cid, writer).catch(console.error)
```

PR #764, opened 2025-03-13
feat: add ipns republishing

2color:
- **feat: add republish record function**
- **fix: linting errors**
- **feat: add republishing logic**


packages/ipns/src/index.ts
```diff
@@ -566,29 +569,70 @@ class DefaultIPNS implements IPNS {
...
+            recordsToRepublish.push({ routingKey, record: ipnsRecord })
+          } catch (err) {
+            this.log.error('error unmarshaling record', err)
+          }
+        }
+
+        this.log(`found ${recordsToRepublish.length} records to republish`)
+
+        // Republish each record
+        for (const { routingKey, record } of recordsToRepublish) {
```
achingbrain, line 598: We might want to do this in a queue so we can publish multiple records in parallel.

achingbrain: The queue implementation we use elsewhere in the stack is [this one](https://github.com/libp2p/js-libp2p/blob/main/packages/utils/src/queue/index.ts), it has simple concurrency controls, etc.

2color:
Following a discussion offline with @achingbrain, we need to move key management internally to implement reproviding. 

Key management is typically (and store encrypted at rest) handled by the js-libp2p keychain, however we don't have libp2p in `@helia/http` (which can also be used for @helia/ipns with delegated routers). This means that we probably need a dedicated keychain instance in the IPNS class, so that we don't depend on the libp2p instance.

achingbrain:
> This means that we probably need a dedicated keychain instance in the IPNS class

You could also use a libp2p instance that's only configured with HTTP routers?  Then you don't need to handle things differently at this level.

2color:
> @helia/http could also use a libp2p instance that's only configured with HTTP routers? Then you don't need to handle things differently at this level.

Yeah we could try that first and see how it feels. It may increase the [bundle size](https://packagephobia.com/result?p=%40helia%2Fhttp). 

I also vaguely remember there being a bug with libp2p whereby instantiating it without the DHT, causes an infinite random walk loop with the defaults, as it's looking for a circuit relay reservation by default. 

Would this also allow Helia to make use of the libp2p over http stuff you've be working on?

achingbrain:
> It may increase the [bundle size](https://packagephobia.com/result?p=%40helia%2Fhttp).

Looking at the breakdown, the usual suspects are already there - `@libp2p/crypto`, `@noble/hashes`, etc so it might not increase it by that much.

> I also vaguely remember there being a bug with libp2p whereby instantiating it without the DHT, causes an infinite random walk loop with the defaults, as it's looking for a circuit relay reservation by default.

I'd imagine this configuration wouldn't configure a relay listener so it wouldn't try to find a relay?

We need this landing, really - https://github.com/ipfs/specs/pull/476

> Would this also allow Helia to make use of the libp2p over http stuff you've be working on?

Yes, though you'd need transports etc in order to use libp2p for the transport layer.

2color:
As discussed in https://github.com/ipfs/helia/discussions/807, the next steps here:
- Take named keys (string), persisted in the keychain using a separate namespace) as input to [`publish`](https://ipfs.github.io/helia/interfaces/_helia_ipns.index.IPNS.html#publish). Keys will be generated internally. 
- Republish will iterate over records in datastore 
- Add `unpublish(‘keyname’)` for deleting records
- Add the `public key` to the return type of `publish`, because we can't derive the name when the key is managed by the keychain.

2color:
I have an open question: we currently store the marshalled DHT record for IPNS records in the datastore. 

As part of republishing, we need to also persist some additional local metadata associated with the IPNS record, like the keychain’s keyname (the key name string that the user passes) associated with the record and potentially also the lifetime. Both of these are needed for republishing. 

The two ways I’ve explored storing this metadata are: 
- Take the current libp2p DHT record we store and make it part of an envelope which also stores some local metadata (lifetime and keychain’s keyname)
- Store the local metadata using a separate key in the datastore specific for ipns metadata. 

The latter approach seems a bit easier and less likely to have downstream effects or unintended consequences (like polluting `/dht/record/` namespaced keys in the datastore with other data that isn’t part of a DHT record)

Since the datastore expect binary values, what are the conventions around storing such data in binary format? Should I use Protobuf for this or can I use cbor (which we already depend on)

For the sake of the example, let's assume I have the following struct that I want to persiste:
```
export interface IPNSRecordMetadata {
  keyName: string
  lifetime: number
}
```

@achingbrain

packages/ipns/src/routing/local-store.ts
```diff
@@ -1,42 +1,63 @@
...
-  put(routingKey: Uint8Array, marshaledRecord: Uint8Array, options?: PutOptions): Promise<void>
+  /**
+   * Put an IPNS record into the datastore
+   *
+   * @param routingKey - The routing key for the IPNS record
+   * @param marshaledRecord - The marshaled IPNS record
+   * @param metadata - local publishing metadata for the IPNS record (optional)
+   * @param options - options for the put operation (optional)
+   */
+  put(routingKey: Uint8Array, marshaledRecord: Uint8Array, metadata?: IPNSPublishMetadata, options?: PutOptions): Promise<void>
```
2color, line 43: Move the metadata to the options.

packages/ipns/src/index.ts
```diff
@@ -782,7 +892,9 @@ class DefaultIPNS implements IPNS {
 
       await ipnsValidator(routingKey, marshaledRecord) // validate that they key corresponds to the record
 
-      await this.localStore.put(routingKey, marshaledRecord, options) // add to local store
+      // TODO: If we are republishing a signed record without access to the key
+      // we can probably skip storing it in the local store
+      // await this.localStore.put(routingKey, marshaledRecord, undefined, options) // add to local store
```
SgtPooki, line 897: this still need resolved.

packages/ipns/src/index.ts
```diff
@@ -568,33 +604,107 @@ class DefaultIPNS implements IPNS {
...
+        for await (const { routingKey, record, metadata } of this.localStore.list({
+          signal: options.signal,
+          onProgress: options.onProgress
+        })) {
+          try {
+            if (metadata == null) {
+              // Skip if no metadata is found from before we started
+              // storing metadata or for records republished without a key
+              this.log(`no metadata found for record ${routingKey.toString()}, skipping`)
+              continue
```
SgtPooki, line 628: reminder: might be worth calling out in release notes or PR that previously published records will need to be published again explicitly in order for republish to work.

packages/ipns/src/index.ts
```diff
@@ -568,33 +604,107 @@ class DefaultIPNS implements IPNS {
...
+      this.timeout = setTimeout(() => {
+        republishRecords().catch(err => {
+          this.log.error('error republishing', err)
         })
       }, nextInterval)
     }
 
+    // TODO: Should we kick off the republish immediately when called?
+    // Queue the first republish.
     this.timeout = setTimeout(() => {
```
SgtPooki, line 695: Should we clear this timeout on the `stop` or "shutdown" of the node? Currently, the timeout is only cleared if the signal passed to republish is aborted.

If a user doesn't abort that signal and the Helia node is stopped, we'll have a timeout leak. Granted, a user should abort the signal if they stop the node, but it would be nice to prevent potential issues for users.

packages/ipns/src/routing/local-store.ts
```diff
@@ -90,7 +117,52 @@ export function localStore (datastore: Datastore): LocalStore {
...
+            const routingKeyBase32 = keyString.substring(DHT_RECORD_PREFIX.length)
+            const routingKey = uint8ArrayFromString(routingKeyBase32, 'base32')
+
+            const metadataKey = ipnsMetadataKey(routingKey)
+            let metadata: IPNSPublishMetadata | undefined
+            try {
+              const metadataBuf = await datastore.get(metadataKey, options)
+              metadata = IPNSPublishMetadata.decode(metadataBuf)
+            } catch (err: any) {
+              log.error('Error deserializing metadata for', routingKeyBase32, err)
```
SgtPooki, line 148:
```suggestion
              log.error('Error deserializing metadata for %s - %e', routingKeyBase32, err)
```

packages/ipns/src/routing/local-store.ts
```diff
@@ -90,7 +117,52 @@ export function localStore (datastore: Datastore): LocalStore {
...
+
+            yield {
+              routingKey,
+              metadata,
+              record: libp2pRecord.value,
+              created: libp2pRecord.timeReceived
+            }
+          } catch (err) {
+            // Skip invalid records
+            log.error('Error deserializing record:', err)
```
SgtPooki, line 159:
```suggestion
            log.error('Error deserializing record - %e', err)
```

packages/ipns/test/resolve-dnslink.spec.ts
```diff
@@ -200,83 +184,4 @@ describe('resolveDNSLink', () => {
     expect(result.cid.toString()).to.equal(cid.toV1().toString())
     expect(result.path).to.equal('foobar/path/123')
   })
-
-  it('should follow CNAMES to delegated DNSLink domains', async () => {
```
SgtPooki, line 204: curious why these tests are removed? Do we no longer support this, or do we feel that they're unnecessary? If the latter, I would rather see a separate PR for this.

packages/ipns/test/republish.spec.ts
```diff
@@ -1,107 +1,481 @@
...
+        await new Promise(resolve => setTimeout(resolve, 2))
+        // Should not republish due to keychain error (key not found)
+        expect(putStubCustom.called).to.be.false()
+        expect(putStubHelia.called).to.be.false()
+      })
+
+      it('should handle datastore errors', async () => {
+      // This test is harder to implement with real datastore since we can't easily
+      // make the datastore fail. Instead, we'll test that the function handles
+      // empty datastore gracefully
```
SgtPooki, line 468: FYI we should be able to stub datastore calls to throw Errors if we feel strongly about testing those failures.

packages/ipns/test/republish.spec.ts
```diff
@@ -1,107 +1,481 @@
...
+        resolve()
+      } else {
+        setTimeout(check, 1)
+      }
+    }
+    check()
+  })
+}
+
+describe('republish', () => {
```
SgtPooki, line 27: it would be nice to have a test around how republishing is supposed to work when the helia node passed to the ipns module is stopped.

packages/ipns/src/index.ts
```diff
@@ -559,6 +584,17 @@ class DefaultIPNS implements IPNS {
...
+      return await this.keychain.exportKey(keyName)
+    } catch (err: any) {
+      // If no named key found in keychain, generate and import
+      const privKey = await generateKeyPair('Ed25519')
+      await this.keychain.importKey(keyName, privKey)
+      return privKey
+    }
+  }
+
   republish (options: RepublishOptions = {}): void {
```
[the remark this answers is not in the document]
2color: Fixed in 3f84f75605f3337c09ee17e89c68e25e8539f393

packages/ipns/test/republish.spec.ts
2color: I don't feel strongly about testing those failures.

packages/ipns/src/index.ts
2color: I updated the PR description to mention this under breaking changes. 

I'm not sure if that's enough for it to make it into the release notes, or it needs to actually be in the commit message. 

@achingbrain

packages/ipns/src/index.ts
2color: Good point. I don't think we have a mechanism for this in place for submodules of Helia.

Direct components of Helia use the `Startable` interface, like the routing class: https://github.com/ipfs/helia/blob/main/packages/utils/src/routing.ts#L55

But in this case, the dependency is inverted, making it a bit trickier. I'll think about this a bit.

packages/ipns/test/republish.spec.ts
2color: Timeout clean up is handled in https://github.com/ipfs/helia/pull/764/commits/7fc3971f246763392b8d5b40d88bed0bd97f843c

packages/ipns/test/republish.spec.ts
2color: In 1da8b6315a81f2cb4213a591ffadf864c669cc27 and 72c079c8f8ea1140bf38f80f7cb9f1b6d2553dee I added some tests to handle localstore (wrapper around the datastore) errors.

packages/ipns/src/index.ts
```diff
@@ -568,33 +630,117 @@ class DefaultIPNS implements IPNS {
...
-          log.error('error republishing', err)
+      // Queue the next republish
+      this.timeout = setTimeout(() => {
+        republishRecords().catch(err => {
+          this.log.error('error republishing', err)
         })
       }, nextInterval)
     }
 
+    // TODO: Should we kick off the republish immediately when called?
```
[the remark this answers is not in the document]
2color: The only problem is that if you publish a record and than start the repbulish loop, it will immediately try to republish a new version of the record. 

To fix this, we need to check when the last time a record was published and its expiry and based on that decide whether to republish.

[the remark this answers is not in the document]
2color: Fixed in https://github.com/ipfs/helia/pull/764/commits/25082d16f5563cadf045bafcaefb1b8acb746dc0

PR #767, opened 2025-03-14
feat: car export supports custom dag traversal and export

SgtPooki:
## Title

feat: support dag-scope

## Description

This PR adds support for the `dag-scope` query parameter to the export and stream methods. This parameter allows the user to specify the scope of the DAG they want to retrieve. The possible values are `all`, `entity`, and `block`. The default value is `all`.

Also refactored the code quite a bit to support different strategies for walking the dag, and added two new options: dagRoot & knownDagPath.

This code also updates the `blockFilter` check to occur before the `blockstore.get` call to prevent potentially unnecessary network calls.

Related https://github.com/ipfs/helia-verified-fetch/pull/198
Related https://github.com/ipfs/service-worker-gateway/issues/356

## Change checklist


packages/car/src/dag-scope.ts
```diff
@@ -0,0 +1,27 @@
+/**
```
SgtPooki, line 1: I don't like that this is in it's own file.. but i was trying to avoid a circular dep, and it's not quite a "type" so...

packages/car/src/index.ts
```diff
@@ -165,10 +158,10 @@ export interface Car {
...
    */
-  stream(root: CID | CID[], options?: AbortOptions & ProgressOptions<GetBlockProgressEvents>): AsyncGenerator<Uint8Array>
+  stream(root: CID | CID[], options?: ExportCarOptions): AsyncGenerator<Uint8Array, void, undefined>
 }
 
-const DAG_WALK_QUEUE_CONCURRENCY = 1
+const log = logger('helia:car')
 
 class DefaultCar implements Car {
   private readonly components: CarComponents
```
SgtPooki, line 167: we should probably modify CarComponents to accept an optional ComponentLogger

packages/car/src/index.ts
```diff
@@ -200,19 +193,71 @@ class DefaultCar implements Car {
...
+      } else if (!options.dagRoot.equals(knownPath[0])) {
+        throw new Error('knownDagPath must start with dagRoot')
+      }
+
+      // Ensure the last CID in the path is one of the target roots
+      const lastCid = knownPath[knownPath.length - 1]
+      const isTargetRoot = roots.some(r => r.equals(lastCid))
+      if (!isTargetRoot) {
+        throw new Error('knownDagPath must end with one of the target roots')
+      }
```
SgtPooki, line 212: we need some solution here to handle multiple roots where there are paths to each root.

I know this sounds confusing, but "root" according to current kubo/boxo is the "target" root block of a dag.. the "dagRoot" is the root of the dag walked to get to that subdag starting node

packages/car/src/index.ts
```diff
@@ -200,19 +193,71 @@ class DefaultCar implements Car {
...
+    if (options?.dagRoot != null && options?.knownDagPath != null && options.knownDagPath.length > 0) {
+      const knownPath = options.knownDagPath
+      void queue.add(async () => {
+        await this.#processKnownPathStrategy(knownPath, queue, writer, options)
+      })
+        .catch(() => {})
+    } else if (options?.dagRoot != null) {
+      void queue.add(async () => {
+        const dagRoot = options.dagRoot
+        if (dagRoot != null) {
```
SgtPooki, line 234: this should be cleaned up.. we don't need to check dagRoot twice.

packages/car/test/fixtures/car-equals.ts
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Car files are identical if:
+ * 1. They have the same root(s)
+ * 2. They have the same blocks
+ * 3. The blocks are the same
+ * 4. The block order is the same
+ * 5. The codec is the same
+ * 6. The car version is the same
+ * 7. The car header is the same
+ * 8. The car length is the same
```
SgtPooki, line 25:
```suggestion
```

SgtPooki (commented):
self review

SgtPooki:
cspell is yelling about some words in dag-scope.spec.ts

SgtPooki:
ping @achingbrain @2color I think some other eyes on this would be good, and this is blocking https://github.com/ipfs/service-worker-gateway/issues/356

packages/car/src/index.ts
```diff
@@ -236,23 +240,79 @@ class DefaultCar implements Car {
...
+      if (!isTargetRoot) {
+        throw new Error('knownDagPath must end with one of the target roots')
+      }
+
+      // knownDagPath is valid, we should double-check that the blockstore has all the blocks
+      const blockstore = this.components.blockstore
+      for (let i = 0; i < knownPath.length; i++) {
+        const cid = knownPath[i]
+        if (!(await blockstore.has(cid))) {
+          throw new Error(`CID in knownDagPath at index ${i} not found in blockstore`)
```
achingbrain, line 267: Is this correct?  What if we want to pull the block from the network?

If the user doesn't want network activity they should be able to pass an `offline` flag which the networked blockstore will use to throw an error when a block is missing.

packages/car/src/types.ts
```diff
@@ -0,0 +1,75 @@
...
+   * An ordered array of CIDs representing the known path from dagRoot to the target root.
+   * The array should start with dagRoot (at index 0) and end with the target root CID.
+   *
+   * If you provide `knownDagPath`, you should already have verified that the CIDs in the path are present in the blockstore.
+   *
+   * This allows optimizing the path traversal by skipping the expensive path-finding process
+   * when the path from dagRoot to the target root is already known, reducing network requests
+   * and computation time.
+   *
+   * Example usage:
```
achingbrain, line 66: There's a tag for this:

```suggestion
   * @example
   *
```

packages/car/src/types.ts
```diff
@@ -0,0 +1,75 @@
...
+   *
+   * If you provide `knownDagPath` and no `dagRoot`, the first CID in `knownDagPath` will be used as `dagRoot`.
+   */
+  dagRoot?: CID
+
+  /**
+   * An ordered array of CIDs representing the known path from dagRoot to the target root.
+   * The array should start with dagRoot (at index 0) and end with the target root CID.
+   *
+   * If you provide `knownDagPath`, you should already have verified that the CIDs in the path are present in the blockstore.
```
achingbrain, line 60: Does it hurt to let the user control this with an `offline` flag?

packages/car/src/dag-scope.ts
achingbrain: You could make it a type and declare it in the index file?

```
type DagScope = 'block' | 'entity' | 'all'
```

Using types instead of enums means we won't need to use [--experimental-transform-types](https://nodejs.org/docs/latest-v23.x/api/cli.html#--experimental-transform-types) in future.

packages/car/src/index.ts
```diff
@@ -236,23 +240,79 @@ class DefaultCar implements Car {
...
+      const decodedBlock = createUnsafe({ bytes, cid, codec })
 
-    const block = createUnsafe({ bytes, cid, codec })
+      // Determine if we should traverse using the provided strategy
+      if (strategy.shouldTraverse(cid, options)) {
+        this.log.trace('traversing cid %s with strategy %s', cid.toString(), strategy.constructor.name)
+        // Process links according to the strategy
+        for await (const result of strategy.getNextCidStrategy(cid, decodedBlock)) {
+          const blockCid = (result as { cid: CID }).cid ?? result
+          const newStrategy = (result as { strategy: TraversalStrategy }).strategy ?? strategy
```
achingbrain, line 306: Can we avoid the casts here?  We don't get type safety like this.

achingbrain (commented):
Initial pass, need to look in more detail tomorrow

packages/car/src/types.ts
SgtPooki: I think the offline flag could do what we want. I will update this and the test

packages/car/src/index.ts
SgtPooki: exportCarOptions doesn't currently accept the offline flag, I will add it

SgtPooki:
@achingbrain addressed all your comments

achingbrain (changes requested):
Just trying to think about the API with fresh eyes.

As I see it there are a few use cases we are trying to cover here.

## Terminology

Root CID - the root of a DAG, e.g. the starting point for traversal.  Has no ascendants, only descendants.  Every node in the DAG can be traversed to starting at this node.

Target CID - We wish to terminate traversal at this node and either export the block it refers to, or the subgraph (or part of?) underneath it.  Must be the root CID or a child of the root CID.

## Use cases

<ol>
<li>Pass a root CID, export the whole DAG beneath it</li>
<li>Pass a root CID and a target CID
  <ol>
    <li>Search the DAG for the target CID starting at the root
      <ol>
        <li>Export the whole subgraph under the target CID</li>
        <li>Export the target CID block</li>
      </ol>
    </li>
    <li>Traverse a predetermined path through the DAG to the target CID starting at the root
      <ol>
        <li>Export the whole subgraph under the target CID</li>
        <li>Export the target CID block</li>
      </ol>
    </li>
  </ol>
</li>
</ol>

Also optionally include the blocks from the path from the root CID to the target CID, if they are not the same.

## Examples

```ts
// Case 1 - export everything under the root
await c.export(rootCid, writer)
```

```ts
// Case 2ia - search the graph under the root and export everything under the target
await c.export(rootCid, writer, {
  traversal: new GraphSearch(targetCid),
  exporter: new SubgraphExporter(),
  includePath?: boolean
})
```

```ts
// Case 2ib - search the graph for the target, only export the target block
await c.export(rootCid, writer, {
  traversal: new GraphSearch(targetCid),
  exporter: new BlockExporter(),
  includePath?: boolean
})
```

```ts
// Case 3ia - walk the path to the target, export everything underneath the target
await c.export(rootCid, writer, {
  traversal: new PathStrategy([rootCid ...intermediateCids, targetCid]),
  exporter: new SubgraphExporter(),
  includePath?: boolean
})
```

```ts
// Case 3ib - walk the path to the target, only export the target block
await c.export(rootCid, writer, {
  traversal: new PathStrategy([rootCid ...intermediateCids, targetCid]),
  exporter: new BlockExporter(),
  includePath?: boolean
})
```

## Traversal strategies

### GraphSearch

A traversal strategy that performs a breadth-first search (so as to not load blocks unnecessarily) looking for a target CID. Traversal stops when we reach the target CID or run out of nodes.

### PathStrategy

A traversal strategy that accepts a list of CIDs and follows the list through the DAG in a depth-first fashion.

## Export strategies

When traversal has ended the export begins starting at the target CID.

### BlockExporter

Yields the first block from the first CID and stops.

### SubgraphExporter

Traverses the DAG depth-first starting at the target CID and yields all encountered blocks.

Can use parallelism to speed up loading blocks but should yield them in inorder order.

The target CID could be a UnixFS file, in which case the CAR would hold just that file, or it could be a UnixFS directory in which case the CAR would hold the whole directory.

It could be a DAG-CBOR block in which case the CAR would hold the root CBOR block and all linked blocks.

### Others

- GlobExporter? (e.g. UnixFS directory traversal, only exports files matching a globstar, e.g. something like `**/*.txt`)
- FilterExporter?  (e.g. a partial sync of a DAG based on CIDs present or absent from a [Filter](https://libp2p.github.io/js-libp2p/interfaces/_libp2p_utils.filters.Filter.html))

-----

With all that in mind we can fully specify case 1:

```ts
// Case 1, specified fully
await c.export(rootCid, writer, {
  // start searching at the root, good news - it's the one we want!
  traversal: new GraphSearch(rootCid),
  // export everything under the root
  exporter: new SubgraphExporter(),
  // could also be true, wouldn't make any difference here
  includePath: false
})
```

packages/car/src/index.ts
```diff
@@ -236,23 +239,72 @@ class DefaultCar implements Car {
...
+        for await (const result of strategy.getNextCidStrategy(cid, decodedBlock)) {
+          void queue.add(async () => {
+            await this.#processBlock(result.cid, queue, writer, result.strategy, options)
+          })
+        }
+      }
+    } catch (err: any) {
+      if (err.name === 'NotFoundError') {
+        this.log?.error('block %c not found in blockstore', cid)
+        throw new BlockNotFoundError(cid)
```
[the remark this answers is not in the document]
SgtPooki: removed

packages/car/src/types.ts
```diff
@@ -0,0 +1,98 @@
+import type { CodecLoader } from '@helia/interface'
+import type { GetBlockProgressEvents } from '@helia/interface/blocks'
+import type { AbortOptions, ComponentLogger } from '@libp2p/interface'
+import type { Filter } from '@libp2p/utils/filters'
+import type { Blockstore } from 'interface-blockstore'
+import type { CID } from 'multiformats/cid'
+import type { ProgressOptions } from 'progress-events'
+
+export interface CarComponents {
```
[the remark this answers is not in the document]
SgtPooki: those types were still being exported from index.ts, but I moved them back in there. I left the unexported strategy/traversal types in `types.ts`

packages/car/src/types.ts
```diff
@@ -0,0 +1,98 @@
...
+   * query, after blocks required to verify path segments
+   *
+   * 'block' - Only the root block at the end of the path is returned after blocks required
+   * to verify the specified path segments.
+   *
+   * 'entity' - For queries that traverse UnixFS data, 'entity' roughly means return blocks
+   * needed to verify the terminating element of the requested content path.
+   * For UnixFS, all the blocks needed to read an entire UnixFS file, or enumerate
+   * a UnixFS directory. For all queries that reference non-UnixFS data, 'entity'
+   * is equivalent to 'block'
```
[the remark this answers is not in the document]
SgtPooki: this is straight from the dag-scope spec: https://specs.ipfs.tech/http-gateways/trustless-gateway/#dag-scope-request-query-parameter

SgtPooki:
I'm refactoring these code changes to follow the API you have set out. I think that will be a little more flexible and from what you laid out, it seems like it will handle all of the scenarios we need (and more)

SgtPooki:
Ok @achingbrain I've got things refactored to use the traversal and export strategy API you suggested. 

One thing I don't like about these changes is that we have to do traversal twice, but if root===target or target===some-leaf-node, it's practically no extra work on either side.

Let me know if you see anything that can be improved.

SgtPooki (commented):
self review

packages/car/src/car.ts
```diff
@@ -0,0 +1,234 @@
...
+    const codec = await this.components.getCodec(cid.code)
+    const bytes = await this.components.blockstore.get(cid, options)
+
+    if (writer != null) {
+      // writer is not null, so we are writing to a car file
+      // Mark as processed
+      options?.blockFilter?.add(cid.multihash.bytes)
+
+      // Write to CAR
+      await writer.put({ cid, bytes })
```
SgtPooki, line 182: I don't think we need any of the writer stuff in here (`traverseDagNode`) anymore

SgtPooki (commented):
some minor nits and cleanup could be done, but makes sense and mostly lgtm

Issue #771, opened 2025-04-07
Allow disabling provider randomization from `HTTPGatewayRouter`


m0ar:
Hi 👋 

I'd like to request a way to disable the forced provider order randomization in `HTTPGatewayRouter`. I've debugged spotty gateway retrievals for a while, and found the source to be this in `findProviders`.

## Background
Some content that the service I'm writing needs to retrieve and pin can only be fetched from ipfs.desci.com (which has `noFetch: true`, and hence fails fast), and a disjoint set is available from pub.desci.com (which doesn't, and times out).

I configured Helia routing with this `HTTPGatewayRouter`, expecting it to try them in order. This would always be optimal in this use case.
```ts
httpGatewayRouting({
  gateways: ["https://ipfs.desci.com", "https://pub.desci.com"],
}),
```

## Issue
However, there is a "mandatory" randomization in `HTTPGatewayRouter` that shuffles the providers every time, which means that only half of my requests first check the fast gateway, and the other half only after the other gateway times out looking for content only available over the other gateway. 

```ts
async * findProviders (cid: CID<unknown, number, number, Version>, options?: RoutingOptions | undefined): AsyncIterable<Provider> {
  // TODO: opt out of this random sort
  yield * this.gateways.toSorted(() => Math.random() > 0.5 ? 1 : -1).map(info => ({
    ...info,
    protocols: ['transport-ipfs-gateway-http']
  }))
}
```

## Suggested fix
I would like to request the option of passing args down to opt out from this randomization, to enable working with gateway tiers like in our use case.

PR #772, opened 2025-04-07
feat: add provider shuffle option to HTTPGatewayRouter

m0ar:
## feat: allow opting out of provider shuffle in HTTPGatewayRouter

See #771 for motivation.

## Description

For use cases where gateway lookup order is important, this PR allows passing a new config param to `HTTPGatewayRouter` which opts out of the shuffling which happens on each call to `findProviders`.

## Notes & open questions

- The new param is optional
- Default value preserves the original behavior

I.e., the feature is backward-compatible.

## Change checklist

### DIFF
--- a/packages/routers/src/http-gateway-routing.ts
+++ b/packages/routers/src/http-gateway-routing.ts
@@ -17,7 +17,12 @@ export const DEFAULT_TRUSTLESS_GATEWAYS = [
 
 export interface HTTPGatewayRouterInit {
   gateways?: Array<URL | string>
-  skipProviderShuffle?: boolean
+  /**
+   * Whether to shuffle the list of gateways
+   *
+   * @default true
+   */
+  shuffle?: boolean
 }
 
 // this value is from https://github.com/multiformats/multicodec/blob/master/table.csv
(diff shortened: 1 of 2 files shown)


m0ar:
Friendly ping @SgtPooki @achingbrain :cupcake: :eyes:

packages/routers/src/http-gateway-routing.ts
```diff
@@ -17,6 +17,7 @@ export const DEFAULT_TRUSTLESS_GATEWAYS = [
 
 export interface HTTPGatewayRouterInit {
   gateways?: Array<URL | string>
+  skipProviderShuffle?: boolean
```
SgtPooki, line 20: maybe something more explicit like `keepOrder` would be better?

SgtPooki (commented):
code looks fine to me.. i think the option could have a better name

packages/routers/src/http-gateway-routing.ts
SgtPooki: also, we should probably add a jsdoc with @default tag..

achingbrain: Or even more concise - ‘shuffle’ (default: true)?

m0ar: @SgtPooki @achingbrain Thanks for the review!

Just pushed a commit changing it to `shuffle`, flipped the corresponding logic, and added a jsdoc comment with a `@default` tag :heavy_check_mark:

SgtPooki (approved):

Issue #775, opened 2025-04-16
bug: trustless-gateway session edge-case


SgtPooki:
i'm working on https://github.com/ipfs/service-worker-gateway/issues/674

I've got code ready to go in the service worker, and can sometimes get things to abort properly.. however, one situation keeps coming up: 

If a signal is aborted during `getRawBlock`, the session will stay alive forever:

```
helia:trustless-gateway:session:bafkreig7p6kzwgg4hp3n7wpnnn3kkjmpzxds5rmwhphyueilbzabvyexvq:error error querying provider TrustlessGateway { url: URL { href: 'http://127.0.0.1:55970/', origin: 'http://127.0.0.1:55970', protocol: 'http:', username: '', password: '', host: '127.0.0.1:55970', hostname: '127.0.0.1', port: '55970', pathname: '/', search: '', searchParams: URLSearchParams {}, hash: '' }, log: [Function: debug] { namespace: 'helia:trustless-gateway-block-broker:127.0.0.1', useColors: true, color: 32, extend: [Function: extend], destroy: [Function: destroy], enabled: [Getter/Setter], inspectOpts: {}, error: [Function: debug] { namespace: 'helia:trustless-gateway-block-broker:127.0.0.1:error', useColors: true, color: 166, extend: [Function: extend], destroy: [Function: destroy], enabled: [Getter/Setter], inspectOpts: {} }, trace: [Function: debug] { namespace: 'helia:trustless-gateway-block-broker:127.0.0.1:trace', useColors: true, color: 129, extend: [Function: extend], destroy: [Function: destroy], enabled: [Getter/Setter], inspectOpts: {} }, diff: 0, prev: undefined, curr: 1744836310513 }, transformRequestInit: undefined }, evicting from session Error: fetching raw block for CID bafkreig7p6kzwgg4hp3n7wpnnn3kkjmpzxds5rmwhphyueilbzabvyexvq from gateway http://127.0.0.1:55970/ was aborted
    at TrustlessGateway.getRawBlock (file:///Users/sgtpooki/code/work/ipshipyard/ipfs/helia/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts:146:15)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at TrustlessGatewaySession.queryProvider (file:///Users/sgtpooki/code/work/ipshipyard/ipfs/helia/packages/block-brokers/src/trustless-gateway/session.ts:38:19)
[...]

PR #776, opened 2025-04-16
fix: trustless-gateway sessions handle aborts properly 

SgtPooki:
We originally did not abort when the session was idle because a block may have been found.

This PR changes that so that we have a race condition on signal.abort and foundBlock, but I believe this small window is a better situation than having the session never end (as in #775)

### DIFF
--- a/packages/block-brokers/.aegir.js
+++ b/packages/block-brokers/.aegir.js
@@ -88,6 +88,17 @@ const options = {
         // fails validation
         res.end(Uint8Array.from([0, 1, 2, 1]))
       })
+      badGateway.all('/ipfs/bafkreig7p6kzwgg4hp3n7wpnnn3kkjmpzxds5rmwhphyueilbzabvyexvq', (req, res) => {
+        // wait for Number.MAX_SAFE_INTEGER seconds to simulate a slow response. Do not query this cid unless you plan to abort the request.
+        setTimeout(() => {
+          res.writeHead(200, {
+            'content-type': 'application/octet-stream',
+            'content-length': 4
+          })
+          res.end(Uint8Array.from([0, 1, 2, 1]))
+        }, Number.MAX_SAFE_INTEGER)
+      })
+
       badGateway.all('/ipfs/*', (req, res) => {
         // fails
         res.writeHead(500)
--- a/packages/block-brokers/.aegir.js
+++ b/packages/block-brokers/.aegir.js
@@ -59,7 +59,7 @@ const options = {
           res.end(Uint8Array.from([0, 1, 2, 0]))
         } else {
           // no reason to use this in tests without the delay configured.
-          res.writeHead(500, {
+          res.writeHead(400, {
             'content-type': 'text/plain',
             'content-length': 13
           })
--- a/packages/block-brokers/.aegir.js
+++ b/packages/block-brokers/.aegir.js
@@ -46,6 +46,27 @@ const options = {
         // "hello"
         res.end(Uint8Array.from([104, 101, 108, 108, 111]))
       })
+
+      goodGateway.all('/ipfs/bafkreig7p6kzwgg4hp3n7wpnnn3kkjmpzxds5rmwhphyueilbzabvyexvq', async (req, res) => {
+        // if 'delay' header is set, delay the response
+        const delay = req.headers['delay']
+        if (delay) {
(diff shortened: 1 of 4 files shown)


packages/block-brokers/test/trustless-gateway-sessions.spec.ts
```diff
@@ -18,6 +18,16 @@ interface StubbedTrustlessGatewaySessionComponents {
...
 
+/**
+ * Normalize the abort signal reason string to be the same across Node.js and browsers.
+ */
+const getSignal = (msg: string, delay: number): AbortSignal => {
+  const ctrl = new AbortController()
+  const signal = ctrl.signal
+  setTimeout(() => { ctrl.abort(msg) }, delay)
+  return signal
+}
```
SgtPooki, line 29: we should be able to check `.to.eventually.be.rejected().with.property('name', 'AbortError')` instead

SgtPooki (commented):
self review

packages/block-brokers/.aegir.js
```diff
@@ -46,6 +46,27 @@ const options = {
...
+        if (delay) {
+          await new Promise((resolve) => setTimeout(resolve, delay))
+          res.writeHead(200, {
+            'content-type': 'application/octet-stream',
+            'content-length': 4
+          })
+          res.end(Uint8Array.from([0, 1, 2, 0]))
+        } else {
+          // no reason to use this in tests without the delay configured.
+          res.writeHead(500, {
```
achingbrain, line 62: If a required header is missing, maybe use "bad request" instead of "internal server error"?

```suggestion
          res.writeHead(400, {
```

packages/utils/src/abstract-session.ts
```diff
@@ -96,8 +96,18 @@ export abstract class AbstractSession<Provider, RetrieveBlockProgressEvents exte
...
+      if (foundBlock) {
+        this.log.trace('session idle, found block')
+        // we found the block, nothing else to do here.
+        return
+      }
+
+      if (options.signal?.aborted === true) {
+        // the session is idle, we haven't found the block, and the signal was aborted, so we can reject the promise.
+        // Note that there is a slight race condition here where we find the block and the signal is aborted within ~10s of ms of each other.
+        // In that case, the request will be rejected with the signal reason.
```
achingbrain, line 108: I'm confused by this comment - are you saying that we've found the block but we end up rejecting the promise anyway?  This sounds like incorrect behaviour that should be fixed?

packages/block-brokers/test/trustless-gateway-sessions.spec.ts
```diff
@@ -109,4 +109,63 @@ describe('trustless-gateway sessions', () => {
...
+  it('should not abort the session when the signal is aborted if the block is found', async () => {
+    const cid = CID.parse('bafkreig7p6kzwgg4hp3n7wpnnn3kkjmpzxds5rmwhphyueilbzabvyexvq')
+    const block = Uint8Array.from([0, 1, 2, 0])
+    const session = createTrustlessGatewaySession(components, {
+      allowInsecure: true,
+      allowLocal: true,
+      transformRequestInit: (requestInit) => {
+        requestInit.headers = {
+          // The difference here on my machine is 25ms.. if the difference between finding the block and the signal being aborted is less than 22ms, then the test will fail.
+          delay: '478'
```
achingbrain, line 145: I'm not sure we'll get reliable timing of this resolution in CI and this may add flakiness to the test suite.

If we need to replicate precise internal invocation ordering, it would be better to use sinon stubs instead of making actual HTTP requests.

achingbrain:
> open to other ideas on how to handle this

I think the problem here is we're not rejecting the [deferred promise](https://github.com/ipfs/helia/blob/main/packages/utils/src/abstract-session.ts#L610) when the user aborts the request.

The signal is passed to all async operations, the intention there was for rejections to bubble up to the caller, but we [ignore queue errors](https://github.com/ipfs/helia/blob/main/packages/utils/src/abstract-session.ts#L88) to allow trying other providers so nothing happens when the signal fires.

I've opened https://github.com/ipfs/helia/pull/778 targeting this PR - instead of adding more conditional logic it just rejects the promise and clears the queue if the signal fires.  Let me know what you think.

achingbrain (approved):

Issue #790, opened 2025-05-12
helia/http: configurable MaxBlockSize


lidel:
[trustless-gateway.ts#L92](https://github.com/ipfs/helia/blob/95911757de749bdf8cab7591f0fa5473fd1af9cf/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts#L92) does not seem to have any protection against being tricked into fetching terabytes of data. 

Helia should have a response size check and config option similar to [HTTPRetrieval.MaxBlockSize](https://github.com/ipfs/kubo/blob/master/docs/config.md#httpretrievalmaxblocksize) from Kubo 0.35.0-rc1, where there is implicit limit of 2MiB but user is able to override it.

Implementation-wise, either error, or if its more tricky with streaming, perhaps wrap response stream in something that will truncate responses bigger than limit?
If we verify hashes before doing put to blockstore, truncating should be enough, because the hash of truncated response will no longer match CID and the response.

cc @SgtPooki  @achingbrain  – marking this as a bug, because without any limit, every client is at risk of being tricked into opening malicious CID and streaming a lot of garbage data

- Part of https://github.com/ipshipyard/roadmaps/issues/5
- Part of https://github.com/ipshipyard/roadmaps/issues/9

Related:

- https://github.com/ipfs/helia-verified-fetch/issues/200

PR #791, opened 2025-05-13
fix: trustless gateway returned blocks can be limited

SgtPooki:
- **fix: create utility function for limiting response bytes**
- **fix: trustless gateway returned blocks can be limited**

## Title

fix: trustless gateway returned blocks can be limited

## Description

Adds a new utility function `limitedResponse` that ensures the response body is less than a given byte limit.

This is done by:

1. Validating that the `content-length` header is less than the limit, if not, an error is thrown.
2. Consuming the response body until the limit is reached. If the response body is greater than the limit, an error is thrown.

Fixes https://github.com/ipfs/helia/issues/790

## Change checklist

### DIFF
--- a/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
+++ b/packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
@@ -1,4 +1,5 @@
 import { base64 } from 'multiformats/bases/base64'
+import { limitedResponse } from './utils.js'
 import type { ComponentLogger, Logger } from '@libp2p/interface'
 import type { CID } from 'multiformats/cid'
 
@@ -17,6 +18,17 @@ export interface TransformRequestInit {
 export interface TrustlessGatewayComponents {
   logger: ComponentLogger
   transformRequestInit?: TransformRequestInit
+}
+
+export interface GetRawBlockOptions {
+  signal?: AbortSignal
+
+  /**
+   * The maximum number of bytes to allow when fetching a raw block.
+   *
+   * @default 2097152 (2MiB)
+   */
+  byteLimit?: number
 }
 
 /**
@@ -89,7 +101,7 @@ export class TrustlessGateway {
    * Fetch a raw block from `this.url` following the specification defined at
    * https://specs.ipfs.tech/http-gateways/trustless-gateway/
    */
-  async getRawBlock (cid: CID, signal?: AbortSignal): Promise<Uint8Array> {
+  async getRawBlock (cid: CID, { signal, byteLimit = 2097152 }: GetRawBlockOptions = {}): Promise<Uint8Array> {
     const gwUrl = new URL(this.url.toString())
     gwUrl.pathname = `/ipfs/${cid.toString()}`
 
@@ -130,8 +142,11 @@ export class TrustlessGateway {
             this.#errors++
             throw new Error(`unable to fetch raw block for CID ${cid} from gateway ${this.url}`)
           }
+          // limited Response ensures the body is less than 2MiB (or configurable byteLimit)
+          // see https://github.com/ipfs/helia/issues/790
+          const body = await limitedResponse(res, { signal: innerController.signal, byteLimit })
           this.#successes++
-          return new Uint8Array(await res.arrayBuffer())
+          return body
         })
         this.#pendingResponses.set(blockId, pendingResponse)
       }
--- a/packages/block-brokers/src/trustless-gateway/utils.ts
+++ b/packages/block-brokers/src/trustless-gateway/utils.ts
@@ -62,7 +62,7 @@ export async function * findHttpGatewayProviders (cid: CID, routing: Routing, lo
  * If the response contains a content-length header greater than the limit or the actual bytes returned are greater than
  * the limit, an error is thrown.
  */
-export async function limitedResponse (response: Response, { log, byteLimit }: { log?: Logger, byteLimit: number }): Promise<Uint8Array> {
+export async function limitedResponse (response: Response, { log, byteLimit, signal }: { log?: Logger, byteLimit: number, signal?: AbortSignal }): Promise<Uint8Array> {
   const contentLength = response.headers.get('content-length')
   if (contentLength != null) {
     const contentLengthNumber = parseInt(contentLength, 10)
@@ -88,6 +88,10 @@ export async function limitedResponse (response: Response, { log, byteLimit }: {
 
   try {
     while (true) {
+      if (signal?.aborted === true) {
+        throw new Error('Response body read was aborted')
+      }
+
       const { done, value } = await reader.read()
       if (done) {
         break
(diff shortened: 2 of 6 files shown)


SgtPooki (commented):
self review

packages/block-brokers/src/trustless-gateway/utils.ts
```diff
@@ -55,3 +55,71 @@ export async function * findHttpGatewayProviders (cid: CID, routing: Routing, lo
...
   }
 }
+
+/**
+ * A function that handles ensuring the content-length header and the response body is less than a given byte limit.
+ *
+ * If the response contains a content-length header greater than the limit or the actual bytes returned are greater than
+ * the limit, an error is thrown.
+ */
+export async function limitedResponse (response: Response, { log, byteLimit, signal }: { log?: Logger, byteLimit: number, signal?: AbortSignal }): Promise<Uint8Array> {
```
achingbrain, line 65: If `byteLimit` is not optional, let's not pass it as an option:

```suggestion
export async function limitedResponse (response: Response, byteLimit: number, options?: AbortOptions): Promise<Uint8Array> {
```

Also, I think we can lose the logger here - the caller can log any thrown errors.

packages/block-brokers/src/trustless-gateway/utils.ts
```diff
@@ -55,3 +55,71 @@ export async function * findHttpGatewayProviders (cid: CID, routing: Routing, lo
...
+        // No need to consume body here, as we were streaming and hit the limit
+        throw new Error(`Response body is greater than the limit (${byteLimit}), received ${receivedLength} bytes`)
+      }
+    }
+  } finally {
+    try {
+      await reader.cancel()
+    } catch (err) {
+      log?.error('Error cancelling reader', err)
+    }
```
achingbrain, line 113: Why do we not throw if cancelling the reader fails?

achingbrain, line 113: Do we also need to release the lock on the reader?

achingbrain (changes requested):
LGTM, a few nits inline.

packages/block-brokers/src/trustless-gateway/utils.ts
SgtPooki: @achingbrain A few things AFAIK:

1. if we read through and there are no other errors, there is no further read stream unlocked, so an error is irrelevant? (idk if this is possible)
2. If an error throws, an error thrown from the finally block would override the actual error happening during streaming (network error, maxSize, or other)

Let me know if my thinking is wrong here.

SgtPooki: @achingbrain I was under the assumption that .cancel() also released the lock, but I looked into it and that is not the case.

We don't currently have any other readers who would be reading the stream, but we should clean it up.

SgtPooki: image

packages/block-brokers/src/trustless-gateway/utils.ts
```diff
@@ -55,3 +55,71 @@ export async function * findHttpGatewayProviders (cid: CID, routing: Routing, lo
...
+  }
+
+  const reader = response.body?.getReader()
+  if (reader == null) {
+    // no body to consume if reader is null
+    throw new Error('Response body is not readable')
+  }
+
+  const chunks: Uint8Array[] = []
```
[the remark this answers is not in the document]
SgtPooki: looking into this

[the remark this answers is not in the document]
SgtPooki: using uint8arraylist in [`1426550` (#791)](https://github.com/ipfs/helia/pull/791/commits/14265504ce49276a53bd1ae9ffd05eb0dc578f47)

packages/block-brokers/src/trustless-gateway/utils.ts
achingbrain: Fair.  Perhaps we should not await it?

```
reader.cancel()
  .catch(err => {
    log?.error('error cancelling reader - %e', err)
  })
```

N.b. use the `%e` token to include errors in the log output, otherwise the formatting can get all messed up.

Also the convention throughout the codebase is logs start in lower case, messages in thrown errors in sentence case.

SgtPooki:
FYI all of the requested changes have been made. lmk if I missed something. I'll look over all the code again now

packages/block-brokers/test/trustless-gateway-utils.spec.ts
```diff
@@ -51,4 +51,55 @@ describe('trustless-gateway-block-broker-utils', () => {
...
+    })
+
+    await expect(limitedResponse(response, 100)).to.eventually.be.rejected
+      .with.property('message', 'Response body is greater than the limit (100), received 1000000 bytes.')
+  })
+
+  it('limitedResponse handles aborted signals', async function () {
+    const abortController = new AbortController()
+    let pullCount = 0
+    const textEncoder = new TextEncoder()
```
SgtPooki, line 80: uint8arraylist required actual Uint8Array data, so i added textencoder.. should probably just be a Uint8Array directly since we don't care about the content here.

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -19,6 +20,17 @@ export interface TrustlessGatewayComponents {
...
 
+export interface GetRawBlockOptions {
+  signal?: AbortSignal
+
+  /**
+   * The maximum number of bytes to allow when fetching a raw block.
+   *
+   * @default 2_097_152 (2MiB)
+   */
+  maxLimit?: number
```
SgtPooki, line 31:
```suggestion
  maxSize?: number
```
it's maxSize in interface/blocks.. we should normalize.

packages/block-brokers/src/trustless-gateway/utils.ts
```diff
@@ -56,3 +57,69 @@ export async function * findHttpGatewayProviders (cid: CID, routing: Routing, lo
...
+  log?: Logger
+}
+
+/**
+ * A function that handles ensuring the content-length header and the response body is less than a given byte limit.
+ *
+ * If the response contains a content-length header greater than the limit or the actual bytes returned are greater than
+ * the limit, an error is thrown.
+ */
+export async function limitedResponse (response: Response, byteLimit: number, options?: LimitedResponseOptions): Promise<Uint8Array> {
```
SgtPooki, line 72: I left the property name here as `byteLimit` but could change to `maxSize` if we want.. leaving for now

SgtPooki (commented):
self review

packages/block-brokers/src/trustless-gateway/trustless-gateway.ts
```diff
@@ -89,7 +101,7 @@ export class TrustlessGateway {
    * Fetch a raw block from `this.url` following the specification defined at
    * https://specs.ipfs.tech/http-gateways/trustless-gateway/
    */
-  async getRawBlock (cid: CID, signal?: AbortSignal): Promise<Uint8Array> {
+  async getRawBlock (cid: CID, { signal, maxSize = 2_097_152 }: GetRawBlockOptions = {}): Promise<Uint8Array> {
```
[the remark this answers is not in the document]
SgtPooki: fixed with [`23d1c58` (#791)](https://github.com/ipfs/helia/pull/791/commits/23d1c582c0d0c8d0892e7430a8f0341ba10d8497)

PR #815, opened 2025-06-06
fix: abstract session should see provider options for trustless gateway

vasco-santos:
## Change checklist


vasco-santos:
@achingbrain this is what I have observed: https://github.com/ipfs/helia-verified-fetch/issues/251#issue-3124752517

In short, when traversing a DAG, we only have providers for the root. Unfortunately, I can't add a test here at the moment, but you should be able to reproduce this pretty easily with this: https://github.com/vasco-santos/hash-stream-infra-poc-server/pull/9/files

The tests there work with this PR released as a fork, while with official released version it fails

vasco-santos:
@achingbrain just tried in https://github.com/vasco-santos/hash-stream-infra-poc-server/pull/9 and looks like it is working with most recent `@helia/verified-fetch` 🙌🏼 

Thanks for fixing this

PR #817, opened 2025-06-11
docs(@helia/unixfs): example - add browser file/dir

tabcat:
## Title

docs(@helia/unixfs): example - add browser file/dir

## Description

Adds an example using [@cypsela/browser-source](https://github.com/cypsela/browser-source) to add [FileSystemEntry](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemEntry) and [FileSystemHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle) to @helia/unixfs

## Notes & open questions

I haven't tested this fully. Thinking about making a live example.

## Change checklist

### DIFF
--- a/packages/unixfs/src/index.ts
+++ b/packages/unixfs/src/index.ts
@@ -49,7 +49,7 @@
  *
  * Uses [@cypsela/browser-source](https://github.com/cypsela/browser-source) to read [FileSystemEntry](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemEntry) and [FileSystemHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle) files and directories.
  *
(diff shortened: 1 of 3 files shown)


packages/unixfs/src/index.ts
```diff
@@ -44,6 +44,35 @@
  *   console.info(entry)
  * }
  * ```
+ *
+ * @example Adding files and directories in the browser
+ *
+ * Uses [@cypsela/browser-source](https://github.com/cypsela/browser-source) to read [FileSystemEntry](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemEntry) and [FileSystemHandle](https://developer.mozilla.org/en-US/docs/Web/API/FileSystemHandle) files and directories.
+ *
+ * Instances of these data types are available from drag and drop events and window methods like (showOpenFilePicker)[https://developer.mozilla.org/en-US/docs/Web/API/Window/showOpenFilePicker].
```
tabcat, line 52: I wrote this link wrong. fixing

2color:
Looking better now

achingbrain (approved):

PR #820, opened 2025-06-15
fix: handle alredy pinned cid

### LINKED ISSUE #642: Helia throws when adding a block if it is already pinned
When adding content to Helia and setting the 'pin' option to true, if a pin exists the add method with throw a generic error with the message 'Already Pinned'.

The throw is here:
https://github.com/ipfs/helia/blob/main/packages/utils/src/pins.ts#L75

A couple of thoughts:
1) If the content already exists and/or is already pinned, is this really an error? The goal of the client is to have the content pinned, if it is _already_ pinned, the client desire is satisfied already, not unsatisfiable.
2) If an error _must_ be thrown, providing a specific error type or enumeration that could be reliably checked against would be helpful. Checking against an arbitrary message string seems fragile.

Haxry:
## Title
closes #642 

## Change checklist

### DIFF
--- a/packages/utils/src/pins.ts
+++ b/packages/utils/src/pins.ts
@@ -75,7 +75,8 @@ export class PinsImpl implements Pins {
     const pinKey = toDSKey(cid)
 
     if (await this.datastore.has(pinKey)) {
-      throw new Error('Already pinned')
+      yield cid
+      return
     }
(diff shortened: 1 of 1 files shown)


achingbrain:
Closing as abandoned.

PR #822, opened 2025-06-20
feat: use libp2p in @helia/http and move libp2p to the base helia class 

2color:
## What

- This introduces a libp2p instance to @helia/http which is configured such that it only makes HTTP requests. 
- This removes the `HeliaP2P` class and moves the libp2p instance to the base Helia class which both Helia and @helia/http instantiate

## Notes & open questions

- I can't seem to fix all the type errors. They're breaking my brain right now.

## Change checklist


achingbrain:
Any reason not to target main with this PR? It would make #764 less noisy.

2color:
> Any reason not to target main with this PR? It would make https://github.com/ipfs/helia/pull/764 less noisy.Any reason not to target main with this PR? It would make https://github.com/ipfs/helia/pull/764 less noisy.

Yeah let's do that. I'll open another PR.

packages/utils/src/index.ts
```diff
@@ -242,6 +271,7 @@ export class Helia implements HeliaInterface {
       this.datastore,
       this.routing
     )
+    await this.libp2p.start()
```
achingbrain, line 274: `this.libp2p` can be passed to `start` same as the blockstore/datastore/routing implementations.

packages/utils/src/index.ts
```diff
@@ -250,6 +280,7 @@ export class Helia implements HeliaInterface {
       this.datastore,
       this.routing
     )
+    await this.libp2p.stop()
```
achingbrain, line 283: `this.libp2p` can be passed to `stop` same as the blockstore/datastore/routing implementations.

packages/ipns/src/routing/local-store.ts
```diff
@@ -40,10 +40,10 @@ export interface LocalStore {
    *
    * @param routingKey - The routing key for the IPNS record
    * @param marshaledRecord - The marshaled IPNS record
-   * @param keyName - The keyName that was used for storng the key in the keychain
-   * @param options - The options for the put operation
+   * @param metadata - local publishing metadata for the IPNS record (optional)
+   * @param options - options for the put operation (optional)
    */
-  put(routingKey: Uint8Array, marshaledRecord: Uint8Array, metadata: IPNSMetadata, options?: PutOptions): Promise<void>
+  put(routingKey: Uint8Array, marshaledRecord: Uint8Array, metadata?: IPNSPublishMetadata, options?: PutOptions): Promise<void>
```
achingbrain, line 46: If this is optional, move it into the options object for consistency.

PR #830, opened 2025-07-17
feat: add active probing to HTTPGatewayRouter

2color:
## Background

Arkadiy from the Internet Archive mentioned he wanted more control over the default HTTPGatewayRouter.

This demonstrates how you would add active probing to ensure CORS.

## Description 

This PR adds active CORS probing functionality to the HTTPGatewayRouter to ensure only CORS-compatible gateways are returned as providers. The implementation includes caching of probe results to avoid repeated checks.

Key changes:
- Introduces CORS probing with configurable cache duration
- Transforms gateway storage from simple PeerInfo array to GatewayStatus objects with metadata
- Filters providers based on CORS compatibility before yielding them

### DIFF
--- a/packages/routers/src/http-gateway-routing.ts
+++ b/packages/routers/src/http-gateway-routing.ts
@@ -72,7 +72,11 @@ class HTTPGatewayRouter implements Partial<Routing> {
         method: 'GET',
         mode: 'cors'
       })
-      return response.ok || response.status === 404
+
+      const corsHeaders = response.headers.get('access-control-allow-origin')
+      const hasCors = corsHeaders === '*' || corsHeaders?.includes(window.location.origin)
+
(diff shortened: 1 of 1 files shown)


achingbrain:
@parkan I'm missing a little context here - do we need this in the HTTP Gateway routing?  I ask because the gateways it will return are from a hard-coded list supplied by the user ([defaults](https://github.com/ipfs/helia/blob/main/packages/routers/src/http-gateway-routing.ts#L10-L16)), so if they aren't compatible with the client (due to lack of CORS support), they should be omitted from the list.

Checking CORS support for providers found via the delegated/libp2p routing makes more sense to me as these are arbitrary peers found by searching the network and not from a pre-supplied list?

PR #832, opened 2025-07-21
fix: support truncated hashes based on multihash length

2color:
## Description
- Added logic to handle truncated multihashes during verification in networked storage
- Enhanced hash verification to work with partial hash values
- Added test to cover truncated hash scenario

## Background 

Originally reported in https://ipshipyard.slack.com/archives/C0238P3SU4X/p1753096031053949 

## Change checklist

### DIFF
--- a/packages/utils/src/utils/networked-storage.ts
+++ b/packages/utils/src/utils/networked-storage.ts
@@ -411,6 +411,13 @@ export const getCidBlockVerifierFunction = (cid: CID, hasher: MultihashHasher): 
       hash = res
     }
 
+    if (hash.digest.length > cid.multihash.size) {
+      hash = {
+        ...hash,
+        digest: hash.digest.subarray(0, cid.multihash.size)
+      }
+    }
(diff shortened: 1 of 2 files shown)


achingbrain:
I've opened https://github.com/multiformats/js-multiformats/pull/329 (and https://github.com/multiformats/js-multiformats/issues/328 for discussion) which implements this in multiformats and allows specifying limits on a per-hash basis.

Here we would then just need to do:

```js
const res = hasher.digest(block, {
  truncate: cid.multihash.size
})
```

...and it would throw if the truncation is too long/short/not allowed.

2color:
Closing in favour of https://github.com/multiformats/js-multiformats/pull/329

achingbrain:
https://github.com/multiformats/js-multiformats/pull/329 has shipped so this can be accomplished via this now:

```ts
const res = hasher.digest(block, {
  truncate: cid.multihash.size
})
```

PR #833, opened 2025-07-21
fix: propagate provider option through session and broker to trustles…

### LINKED ISSUE #816: bug: @helia/block-brokers session do not propagate provider option for http trustless gateway
I noticed while trying to use `?provider` query parameter with `@helia/verified-fetch` that it does not get propagated to the block brokers, specifically to the trustless gateway in the same way as when one sets a gateway in the `@helia/verified-fetch` config. Resulting in basically being ignore.

This hack https://github.com/ipfs/helia/pull/815/files made it work for me, but I believe it may exist a better solution that I am not aware at the moment.

Thoughts @achingbrain

nitininhouse:
Ensure the `provider` option is correctly propagated from the API, through the session and broker layers, down to the trustless gateway HTTP request.

Fixes #816

### DIFF
--- a/packages/block-brokers/src/trustless-gateway/broker.ts
+++ b/packages/block-brokers/src/trustless-gateway/broker.ts
@@ -10,7 +10,7 @@ import type { CID } from 'multiformats/cid'
 export interface CreateTrustlessGatewaySessionOptions extends CreateSessionOptions<TrustlessGatewayGetBlockProgressEvents> {
   /**
    * By default we will only connect to peers with HTTPS addresses, pass true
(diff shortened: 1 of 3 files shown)


SgtPooki (commented):
I believe this can be replaced by https://github.com/ipfs/helia/pull/834?

achingbrain (changes requested):
Thanks for opening this.

Please address all comments and add tests to prevent regressions.

It cannot be merged without tests.

achingbrain:
Superseded by #834

Issue #801, opened 2025-05-14
@helia/ipns methods to accept CID encoded IPNS key


tabcat:
The ~`publish`~, `republishRecord`, and `resolve` methods should accept CID instances of Libp2p public keys. The specified [string representation](https://specs.ipfs.tech/ipns/ipns-record/#string-representation) for IPNS names uses CID. This may reduce confusion for devs.

@2color

PR #852, opened 2025-09-21
feat(ipns): add CID support for publish, republishRecord, and resolve

Harshdev098:
Fixes: #801 

Added CID encoded IPNS key in `publish`, `republishRecord` and `resolve` methods

### DIFF
--- a/packages/ipns/src/index.ts
+++ b/packages/ipns/src/index.ts
@@ -309,6 +309,7 @@ const log = logger('helia:ipns')
 
 const MINUTE = 60 * 1000
 const HOUR = 60 * MINUTE
+const LIBP2P_KEY_CODEC = 0x72
 
 const DEFAULT_LIFETIME_MS = 48 * HOUR
 const DEFAULT_REPUBLISH_INTERVAL_MS = 23 * HOUR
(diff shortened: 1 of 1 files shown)


packages/ipns/src/index.ts
```diff
@@ -751,25 +809,51 @@ class DefaultIPNS implements IPNS {
...
+              throw new Error(`Invalid string key: ${e.message}`)
+            }
           }
         } else {
-          mh = key
+          const digest = key as MultihashDigest<0x00 | 0x12>
+          if (![IDENTITY_CODEC, SHA2_256_CODEC].includes(digest.code)) {
+            throw new Error('Unsupported multihash code for IPNS key')
+          }
+          mh = digest
```
tabcat, line 856: The only change needed was to normalize key to multihash if it was a CID.

tabcat (changes requested):
Did not mean to include `publish` in the request for changes.

The original issue had to do with usage of the [delegated routing client](https://github.com/ipfs/helia-delegated-routing-v1-http-api)'s IPNS methods.
The routing key inputs for those methods are not compatible with these methods in @helia/ipns which caused some issues for me.
I'd be interested in knowing why they take so many different types. It may be good to look at the commit history.
I know the delegated router needs the libp2p key as a CID because that is how it is encoded in the url. But these methods only need the multihash.

PR #891, opened 2025-10-30
feat: republish third-party ipns record

### LINKED ISSUE #877: feature request: `@helia/ipns` republish specific existing records
I want to add a republish method to the IPNS interface that would take a routing key and an existing IPNS record and republish it regularly to keep it in the DHT until it expires or IPNS.unpublish is called for that routing key.

Adding the method is needed to specify which external records (a record where the publishing private key is not local) in the localStore should be republished.

The IPNS.republish would do something like:

1. check the routing key matches the record - throw if verification fails
2. check the record expiry - throw if expired
3. immediately publish to routers with option: { republish: true }

> The republish option is read by the localStore to add a separate republish key, like the metadata key, to the datastore along with the Record.

Type Signature:

```ts
(key: CID<unknown, 0x72, 0x00 | 0x12, 1> | PublicKey | MultihashDigest<0x00 | 0x12> | PeerId, record: IPNSRecord): Promise<void>
```

tabcat:
## Title

feat: republish third-party ipns record

## Description

Adds the ability to keep alive an IPNS record this node did not originally
publish, plus the upkeep-policy machinery around it.

New methods on the IPNS interface:

- `import(key, record)`: validate a record and store it locally without touching
  the network. No metadata is written, so the record is served on a GET but
  dormant until `republish` gives it an upkeep policy.
- `republish(key, options)`: republish the record held locally for a key. It
  compares the local record against the network with `ipnsSelector`, throws
  `RecordObsoleteError` if a newer record is already published, otherwise
  persists the chosen upkeep policy and publishes. Best-effort on the network (a
  failing network router does not throw; use onProgress and the logs), but a
  local-store persist failure throws, since that write saves the upkeep policy.
- `unpublish(key, options)`: stop republishing. By default it removes only the
  metadata (the record stays served until it expires); `removeRecord: true` also
  deletes the record.

Upkeep policy (stored per key in the metadata, default `reissue`):

- `reissue`: re-sign the record with a new sequence and validity. Needs the
  signing key, so only `publish` can select it.
- `rebroadcast`: re-put the existing record as-is. The background loop resolves
  the network's latest record first and adopts a newer one instead of pushing a
  stale copy.
- `none`: no upkeep.

Other changes:

- `resolve` skips the local store when `nocache` is set, and honors
  `validate: false` on the network path.
- The validator rejects a record whose embedded (non-identity) public key does
  not hash to its routing key.

The metadata `Upkeep` enum replaces the earlier `refresh` bool; its values are
`reissue` / `rebroadcast` / `none`.

Closes #877

## Notes & open questions

- `publish()` is all-or-nothing (a single router failure fails the publish)
  while `republish()` is best-effort on the network.

## Change checklist


Verity-Freedom:
сс @lidel for awareness
cc @tabcat [for awareness](https://github.com/ipfs/kubo/issues/4435#issuecomment-3549270256)

tabcat:
Turning this back into a draft. There should be a way to create a new record without publishing or republishing to public routers as part of this PR.

packages/ipns/src/index.ts
```diff
@@ -34,7 +34,11 @@
  * value.
  *
  * ```TypeScript
- * import { createHelia } from 'helia'
+ * import { createHelia } from 'helia'export interface IPNSRecordMetadata {
+  keyName: string
+  lifetime: number
+  upkeep: Upkeep
+}
```
achingbrain: Looks like a mistake?

```suggestion
 * import { createHelia } from 'helia'
```

packages/ipns/src/pb/metadata.proto
```diff
@@ -1,9 +1,19 @@
 syntax = "proto3";
 
+enum Upkeep {
+  republish = 0; // Republish the record with an updated ttl (key must be in keychain)
+  refresh = 1; // Republish the record as is (while it is still valid)
+  none = 3; // Disables automated republishing for the record
```
achingbrain: Any reason this is not sequential?

```suggestion
  none = 2; // Disables automated republishing for the record
```

tabcat:
this still needs some work. ill convert to ready when its ready for review.

packages/ipns/src/ipns/republisher.ts
```diff
@@ -135,10 +164,94 @@ export class IPNSRepublisher {
...
+      for (const routingKey of keysToRepublish) {
+        // resolve the latest record
+        let latestRecord: IPNSRecord
+        try {
+          const { record } = await this.resolver.resolve(multihashFromIPNSRoutingKey(routingKey))
+          latestRecord = record
+        } catch (err: any) {
+          this.log.error('unable to find existing record to republish - %e', err)
+          continue
+        }
```
tabcat, line 197: wondering if we should be resolving the latest record for refreshing. this server is set as the republisher for this routingKey. it should be the only one publishing to this key on the network. this could just be an offline resolution.

tabcat:
I'm pretty happy with where this is at now. I've updated the description to give a pretty good idea of whats changed.

I did change the behavior of unpublish to not remove the record by default, just the metadata which sets the upkeep. I think this is the correct behavior because deleting the record loses the record sequence number. I've added an option removeRecord for the case where the user actually wants to remove the record. Not sure if this would be considered a breaking change.

packages/ipns/src/ipns/republisher.ts
```diff
@@ -138,36 +179,140 @@ export class IPNSRepublisher {
...
+    // do not downgrade a record we already hold
+    if (await this.localStore.has(routingKey, options)) {
+      const { record: existingBytes } = await this.localStore.get(routingKey, options)
+      const existing = IPNSEntry.decode(existingBytes)
+
+      // if the record we already have wins the selector, the incoming one is obsolete
+      if (ipnsSelector(routingKey, [entry, existing]) !== 0) {
+        throw new RecordObsoleteError('A newer record is already stored for this key', existing)
+      }
+    }
```
achingbrain, line 312: Better to just try to load an existing record and handle the failure rather than async `has`/`get` as it introduces a race.

```suggestion
    try {
      const { record: existingBytes } = await this.localStore.get(routingKey, options)
      const existing = IPNSEntry.decode(existingBytes)

      // if the record we already have wins the selector, the incoming one is obsolete
      if (ipnsSelector(routingKey, [entry, existing]) !== 0) {
        throw new RecordObsoleteError('A newer record is already stored for this key', existing)
      }
    } catch (err: any) {
      if (err.name !== 'NotFoundError') {
        throw err
      }
    }
```

packages/ipns/src/ipns/republisher.ts
```diff
@@ -138,36 +179,140 @@ export class IPNSRepublisher {
...
+
+    const marshaledRecord = IPNSEntry.encode(localRecord)
+    const metadata = {
+      upkeep: Upkeep[options.upkeep ?? 'rebroadcast']
+    }
+
+    // a publish that fails on every router does not throw; use onProgress to see
+    // which routers worked and watch the logs
+    await this.#publishToRouters(routingKey, marshaledRecord, {
```
achingbrain, line 283: We can skip the re-encode here and just use the record from the local store?

```suggestion
    const metadata = {
      upkeep: Upkeep[options.upkeep ?? 'rebroadcast']
    }

    // a publish that fails on every router does not throw; use onProgress to see
    // which routers worked and watch the logs
    await this.#publishToRouters(routingKey, record, {
```

packages/ipns/src/ipns/republisher.ts
```diff
@@ -138,36 +179,140 @@ export class IPNSRepublisher {
...
+  async republish (key: MultihashDigest, options: RepublishOptions = {}): Promise<RepublishResult> {
+    const routingKey = multihashToIPNSRoutingKey(key)
+
+    // read the record we hold for this key directly, it was validated when it
+    // was imported or published, and reading it through resolve() would recurse
+    // and rewrite the local store
+    if (!(await this.localStore.has(routingKey, options))) {
+      throw new NotFoundError('No local record found to republish - import it first')
+    }
+
```
achingbrain, line 258: If the record is not present `.get` will throw `NotFoundError` so the extra async call adds overhead - unless the more specific error message is important enough to keep, in which case you could wrap the `.get` call and override the error message?

```suggestion
```

packages/ipns/test/republish.spec.ts
```diff
@@ -418,6 +605,413 @@ describe('republish', () => {
...
+        expect(result.heliaRouting.put.called).to.be.false('published expired record')
+      })
+    })
+
+    describe('error handling', () => {
+      it('should throw if there is no local record to republish', async () => {
+        const key = await result.keychain.generateKey('test-key')
+        const routingKey = multihashToIPNSRoutingKey(key.publicKey.toMultihash())
+
+        await expect(name.republish(multihashFromIPNSRoutingKey(routingKey))).to.be.rejectedWith('No local record found to republish')
```
achingbrain, line 865: Assert on the type of error thrown so the messages can be improved without breaking the tests.

```suggestion
        await expect(name.republish(multihashFromIPNSRoutingKey(routingKey))).to.eventually.be.rejected
          .with.property('name', 'NotFoundError')
```

achingbrain (changes requested):
LGTM, a few small nits inline.

packages/ipns/src/ipns/republisher.ts
tabcat: good catch

achingbrain (approved):

PR #906, opened 2025-11-19
feat: ipns libp2p fetch

### LINKED ISSUE #893: IPNS should use libp2p-fetch to retrieve IPNS records from peers
The IPNS-over-pubsub implementation waits for peers to publish updated IPNS records in order to resolve them.

Instead it should query topic peers directly for the record over libp2p-fetch.

This needs to be implemented [here](https://github.com/ipfs/helia/blob/main/packages/ipns/src/routing/pubsub.ts#L145-L166).

Keys should be sent as `Uint8Array`s and not strings to avoid data corruption.

Spec: https://specs.ipfs.tech/ipns/ipns-pubsub-router/#layering-persistence-onto-libp2p-pubsub
Libp2p-fetch: https://libp2p.github.io/js-libp2p/interfaces/_libp2p_fetch.Fetch.html#fetch

tabcat:
## Title

feat: ipns libp2p fetch

## Description

Adds `@libp2p/fetch` functionality to the pubsub router of `helia/ipns`.
The pubsub router will libp2p/fetch the latest local record from new pubsub peers if libp2p.services.fetch is found.

https://specs.ipfs.tech/ipns/ipns-pubsub-router/#layering-persistence-onto-libp2p-pubsub

Closes #893

## Notes & open questions

Should `get` method of the pubsub router wait for a few peers to be queries with fetch, or should the router emit an event when it finds a newer record?

## Change checklist


tabcat:
I'm considering checking the current subscribers for the topic directly before the fetch call. The check would see if we are still pubsub peered with the peerId that is being queried with the fetch call. If not then cancel the fetch and return. This would reduce the possibility of opening a new connection instead of just a new stream.

PR #1013, opened 2026-04-29
feat: add walker option to pins.add and pins.rm

tabcat:
## Title

feat: add walker option to pins.add and pins.rm

## Description

Adds an optional `walker?: GraphWalkerFactory` to `AddOptions` and
`RmOptions`, letting callers pick the traversal strategy used when
walking the DAG to pin/unpin. The default is `depthFirstWalker()`, so
the existing yield order is preserved bit-for-bit and no caller needs
to change.

Motivating use case: balanced fan-out when pinning very wide DAGs from
gateway-style consumers, where the current DFS recursion pegs one
branch's bandwidth and starves siblings. With `breadthFirstWalker()`
(or a user-supplied factory) callers get level-by-level traversal.

While here:

- Adds `depth?: number` to `WalkOptions` so the walker abstraction
  honors `pins`'s existing `depth` argument without a post-walk filter.
  All three concrete walkers (`depthFirstWalker`, `breadthFirstWalker`,
  `naturalOrderWalker`) gate child enqueues on `opts.depth < maxDepth`.
- Moves the walker type contract (`GraphWalker`,
  `GraphWalkerComponents`, `GraphWalkerInit`, `GraphNode`,
  `WalkOptions`, new `GraphWalkerFactory`) from `@helia/utils` to
  `@helia/interface` so `pins.ts` (in `interface`) can reference
  `GraphWalkerFactory`. `@helia/utils` re-exports the same names — no
  public-surface break.
- Deletes `PinsImpl#walkDag` and its `Queue` plumbing. `add()` and
  `rm()` now route through `walker.walk()`.

## Notes & open questions

- Default behavior is unchanged. `walker` and `WalkOptions.depth` are
  new optional fields; existing callers see no difference.
- Type re-locations to `@helia/interface` are mirrored by re-exports
  in `@helia/utils`, so downstream packages that import walker types
  from `@helia/utils` keep working.

## Change checklist


packages/interface/src/index.ts
```diff
@@ -150,5 +150,6 @@ export interface HeliaEvents<T extends Libp2p = Libp2p> {
 
 export * from './blocks.ts'
 export * from './errors.ts'
+export * from './graph-walker.ts'
```
tabcat, line 208: wondering if graph-walker types should come from ./pins

achingbrain (approved):

achingbrain:
Thanks!

PR #1029, opened 2026-05-11
feat: add configurable delay to trustless gateway router

achingbrain:
Allow slowing down "discovery" of trustless gateway routers to give other routing systems a chance to find providers for a given CID.

## Change checklist

### DIFF
--- a/packages/routers/src/http-gateway-routing.ts
+++ b/packages/routers/src/http-gateway-routing.ts
@@ -7,6 +7,7 @@ import { toString as uint8ArrayToString } from 'uint8arrays/to-string'
 import type { Provider, Routing, RoutingOptions } from '@helia/interface'
 import type { PeerInfo } from '@libp2p/interface'
 import type { Version } from 'multiformats'
+import { delay } from './utils/delay.ts'
 
 export const DEFAULT_TRUSTLESS_GATEWAYS = [
   // 2023-10-03: IPNS, Origin, and Block/CAR support from https://ipfs.github.io/public-gateway-checker/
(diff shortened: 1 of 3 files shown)


packages/routers/src/utils/delay.ts
```diff
@@ -0,0 +1,8 @@
+import type { AbortOptions } from "@libp2p/interface";
+import { raceSignal } from 'race-signal'
+
+export function delay (ms: number, options?: AbortOptions): Promise<void> {
+  return raceSignal(new Promise<void>(resolve => {
+    setTimeout(() => resolve(), ms)
+  }), options?.signal)
+}
```
lidel, line 8: q: do we want to explicitly clear timeout, so on abort the event loop is not kept alive unti ms elapses?

```suggestion
export function delay (ms: number, options?: AbortOptions): Promise<void> {
  let handle: ReturnType<typeof setTimeout>
  return raceSignal(new Promise<void>(resolve => {
    handle = setTimeout(() => resolve(), ms)
  }), options?.signal).finally(() => clearTimeout(handle))
}
```

packages/routers/src/http-gateway-routing.ts
```diff
@@ -48,24 +59,26 @@ class HTTPGatewayRouter implements Partial<Routing> {
...
   }
 
   async * findProviders (cid: CID<unknown, number, number, Version>, options?: RoutingOptions | undefined): AsyncIterable<Provider> {
     yield * (this.shuffle
       ? this.gateways.toSorted(() => Math.random() > 0.5 ? 1 : -1)
       : this.gateways
-    ).map(info => {
-      const provider = {
+    ).map(async info => {
+      await delay(this.delay)
```
lidel, line 75: nit: should `options?.signal` be forwarded to `delay`?

```suggestion
      await delay(this.delay, options)
```

lidel (commented):
some drive-by comments

lidel:
Triage note:
- delays are bit blunt tool
- we will take second look from perspective of "capabilities" and fallback being one of them

achingbrain:
Closing in favour of a capabilities-based approach

PR #1040, opened 2026-05-14
fix: skip identify race when peer protocols are already known

Rinse12:
## Problem

`Network#connectTo` awaits `Promise.all([libp2p.dial, raceEvent('peer:identify', filter for BITSWAP_120)])`. `peer:identify` is single-shot per peer per process. If the peer was already identified before bitswap calls `connectTo` (common when the host app dials peers via other subsystems — pubsub, fetch, kad-dht, manual `dial`), the `raceEvent` waits for an event that will never re-fire and `connectTo` hangs indefinitely.

This blocks `findAndConnect`'s find-providers fallback, so `wantBlock` is forced to time out instead of resolving from already-connected bitswap-speaking peers.

## Fix

Before the race, check `libp2p.peerStore.get(peer)`. If the peer is already known and advertises `BITSWAP_120`, return from a plain `dial()` (cheap no-op when already connected). Otherwise fall through to the existing identify-race code path unchanged. Guarded with `isPeerId(peer)` so the existing `PeerId | Multiaddr | Multiaddr[]` signature still works.

## Safety

- `connectTo` callers (`findAndConnect` at network.ts:319 and 330) discard the returned `Connection` and use `connectTo` only for its peer-warmup side effect.
- The actual bitswap stream is opened later in `sendMessage` via `libp2p.dialProtocol(peerId, BITSWAP_120)`, which performs its own protocol negotiation. So a fast-path-returned connection whose bitswap stream is in fact broken is detected and retried by `sendMessage` — no invariant is violated.

## Measured effect

In an IPNS-over-pubsub workload where pubsub-mesh warmup dials the same peers that bitswap later wants providers from, median \`wantBlock\` wall time drops ~17% (3823 → 3192 ms over n=230 fetches). The improvement matches the theoretical save: the identify-race wait on already-identified peers is eliminated.

## Tests

Added two tests in \`packages/bitswap/test/network.spec.ts\`:

- *\"should not wait for peer:identify when the peer is already known to speak bitswap\"* — stubs \`peerStore.get\` to return a peer whose \`protocols\` already contain \`BITSWAP_120\`, calls \`connectTo\`, and asserts the call resolves without subscribing to \`peer:identify\`. Pre-patch this test would hang because no \`peer:identify\` event is dispatched.
- *\"should fall through to the identify race when the peer is not yet in peerStore\"* — stubs \`peerStore.get\` to reject, simulates \`peer:identify\` firing after \`dial\`, and asserts the slow-path subscription is still installed. Guards against the fast-path swallowing the unknown-peer case.

Lint, dep-check, and the full bitswap node test suite (66 tests) all pass.


packages/bitswap/src/network.ts
```diff
@@ -406,6 +406,21 @@ export class Network extends TypedEventEmitter<NetworkEvents> {
 
     options?.onProgress?.(new CustomProgressEvent<PeerId | Multiaddr | Multiaddr[]>('bitswap:dial', peer))
 
+    // Fast path: peer:identify is single-shot per peer, so if the peer was
+    // already identified (e.g. dialed by another subsystem) the raceEvent
+    // below would wait for an event that has already fired. When the peerStore
+    // already lists BITSWAP_120 for this peer we can skip the race entirely.
+    if (isPeerId(peer)) {
+      try {
+        const peerData = await this.libp2p.peerStore.get(peer)
```
achingbrain: Need to pass options into all async calls, otherwise there's no way to cancel the promise (for example when the user wants an operation to be able to time out) and things will appear to hang.

```suggestion
        const peerData = await this.libp2p.peerStore.get(peer, options)
```

packages/bitswap/test/network.spec.ts
```diff
@@ -66,6 +66,73 @@ describe('network', () => {
...
+
+    components.libp2p.peerStore = stubInterface({
+      get: Sinon.stub().rejects(peerStoreError)
+    }) as any
+
+    components.libp2p.dial.callsFake(async () => {
+      // simulate identify firing after dial
+      setTimeout(() => {
+        const call = components.libp2p.addEventListener.getCalls()
+          .find(c => c.args[0] === 'peer:identify')
```
achingbrain, line 113: I think we can skip waiting for identify entirely since we don't do anything with the established connections that are returned from `connectTo`, instead just letting the topology be invoked when the connection is established and/or dialling the protocol directly.  If identify has already run on the connection, we'll send the peer any new CIDs as WANTs as they come in so again, no need to do anything special here.

Issue #1051, opened 2026-06-09
feat: CAR-stream session for @helia/block-brokers (one gateway request per root)


SgtPooki:
## Proposal

The trustless-gateway broker fetches one `?format=raw` request per block. For a many-small-block DAG (a sharded HAMT directory is hundreds of tiny blocks) that is hundreds of round trips. The graph walker's lookahead overlaps them, but the per-request latency still dominates.

A session that issues one `?format=car&dag-scope=all` request per root and serves the walker's per-block `retrieve()` calls from the streamed CAR collapses that to a single request. Blocks stay hash-verified through the session's `validateFn`, and any block the CAR omits or delivers corrupt falls back to a single `?format=raw` fetch.

## Impact

One streaming request instead of N for any CAR-endpoint-backed retrieval: `@helia/verified-fetch` CAR responses and `@helia/car` export over a trustless gateway. In a local bench (249-block sharded DAG, 25ms/request) the single-CAR path ran ~24x faster than per-block, with byte-identical output.

## The design question

`NetworkedStorage.createSession(root, options)` (`@helia/utils` 2.5.2) passes only `options` to `broker.createSession(options)`, so the broker session never receives the root. A CAR-stream session needs it to open the stream. Options, in order of preference:

1. Pass `root` to `broker.createSession(root, options)`. Small signature change, makes root-scoped streaming brokers first-class.
2. Infer the root from the first `retrieve(cid)` (the walker's first get is the root). Works today, but it is implicit coupling.
3. Leave it, and I keep this in my own project.

It would not subclass `AbstractSession` (that is per-block, per-provider, racing); it implements `SessionBlockBroker` directly with a no-op `addPeer`. Want to confirm that shape is acceptable before I open a PR.

Working implementation (as a `BlockSource` in my project, same pump/index/verified-gap-fill logic): https://github.com/SgtPooki/ipfs2foc/blob/main/packages/core/src/car-stream-source.ts . Happy to share the bench too.

PR #1052, opened 2026-06-09
feat: CAR-stream session for trustless-gateway block broker

SgtPooki:
Implements the block-broker side of #1051.

## What changed

An opt-in CAR-stream session for the trustless-gateway broker: `trustlessGateway({ carStreamSessions: true })`. Instead of one `?format=raw` request per block, a session fetches the whole DAG over one `?format=car&dag-scope=all` request to one gateway and serves each block the graph walk asks for from the stream. Blocks the CAR omits, or that fail the per-retrieve `validateFn`, fall back to a single `?format=raw` fetch.

For a many-small-block DAG (a sharded HAMT directory is hundreds of tiny blocks) that is one request instead of hundreds. A local bench (249-block DAG, 25ms/request) measured ~24x.

Default behaviour is unchanged (`carStreamSessions: false`). The session does not extend `AbstractSession`, which models per-block, per-provider racing; it implements `SessionBlockBroker` directly.

## Two design questions (why this is a draft)

1. `createSession` is not passed the root, so the session infers it from the first `retrieve` (the same way `AbstractSession` scopes provider discovery to the first CID). Passing `root` to `broker.createSession(root, options)` would make this first-class.
2. `SessionBlockBroker` has no close hook, so an abandoned stream cannot be cancelled eagerly. Today it is bounded by the buffer cap plus a 60s idle-timeout teardown; a close hook would let it cancel immediately.

I would rather settle those before polishing, since both touch the session contract.

## Notes / risks

- Verification is unchanged: every block passes through the caller's `validateFn` before it is returned, so a truncated or wrong CAR cannot reach the blockstore as a verified block.
- `getCar` streams the whole DAG, so it is not `limitedResponse`-capped like `getRawBlock`. Per-block bytes above the 2 MiB ceiling are dropped, but `@ipld/car` reads each declared block before that check, so an untrusted gateway is an opt-in DoS surface.
- Follow-ups: a `SessionStorage`/`NetworkedStorage` integration test (the new abort-simulation test is a proxy for the `raceBlockRetrievers` cleanup), full progress-event parity on the stream path, and an oversize-block test.

## Verify

`cd packages/block-brokers && npm run test:node` (Node 24). 30 tests; build, lint, and dep-check green.

### DIFF
--- a/.github/dictionary.txt
+++ b/.github/dictionary.txt
@@ -1,3 +1,5 @@
+backpressures
 dont
+dups
 fleek
 republisher
--- a/packages/block-brokers/.aegir.js
+++ b/packages/block-brokers/.aegir.js
@@ -1,5 +1,46 @@
+import { CarWriter } from '@ipld/car'
 import cors from 'cors'
+import { CID } from 'multiformats/cid'
+import * as raw from 'multiformats/codecs/raw'
+import { sha256 } from 'multiformats/hashes/sha2'
 import polka from 'polka'
+
+/**
+ * Fixed raw blocks used by the CAR-stream session tests. The test recomputes
+ * these from the same bytes, so the CIDs match without passing them around.
+ */
+const CAR_BLOCK_BYTES = [
+  Uint8Array.from([1, 2, 3, 4]),
+  Uint8Array.from([5, 6, 7, 8]),
+  Uint8Array.from([9, 10, 11, 12])
+]
+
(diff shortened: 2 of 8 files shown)


SgtPooki:
Note: this is in draft only because i'm not sure if this is a path we want to take at all, or if this would be accepted in helia. I could publish as a helia module others can use if not desired in the helia repo itself.

cc @achingbrain

Issue #1101, opened 2026-08-09
AbstractSession never removes the listeners it adds to each per-request Queue - unbounded growth in browsers


Rinse12:
> [!IMPORTANT]
> **Retracted 2026-08-12 — the mechanism claim below is wrong.** @achingbrain could not
> reproduce it and was right not to. A listener-bearing `EventTarget` *is* collected in Gecko:
> the cycle collector reclaims it, cycle and all. `root kind = "wrapper, holder …"` in a
> `gc-edges` dump is the **JS GC's** view alone and says nothing about cycle-collectability,
> which is what I mistook for proof of retention.
>
> Measured directly with a `FinalizationRegistry` instead of heap dumps — 20,000 objects per
> arm, run twice — a real `Queue` with all three `abstract-session` listeners still attached
> and nothing else referencing it is collected **100%** of the time. Full numbers and harness
> in [this comment](https://github.com/ipfs/helia/issues/1101#issuecomment-5263740859).
>
> **What actually retained the queues was a separate `main-event` bug**
> ([achingbrain/main-event#13](https://github.com/achingbrain/main-event/issues/13)):
> `retrieve()` also registers `peerAddedToSessionListener` on the long-lived session, and that
> closure captures `queue`. Stock `main-event`'s `removeEventListener` made its removal a
> no-op. With `main-event` patched, every queue is collected **whether or not** the change
> proposed here is applied — the reverse of what this issue claimed.
>
> Specifically retracted: "every `Queue` ever created stays a GC root", the **53.5% of the
> live heap** figure, and the **~930 retained queues/hour** rate. All three rest on the same
> bad premise.
>
> What stands: `retrieve()` adds three listeners to a per-request `Queue` and never removes
> them, while removing the other two listeners in the same `finally`. That is a real
> inconsistency worth fixing as hygiene — see #1102 — but it is **not** a leak.

---

## Summary

`AbstractSession.retrieve()` in `@helia/utils` creates one `Queue` per block request and registers three listeners on it, but the `finally` only calls `queue.clear()` — which splices the job array and touches no listeners:

```ts
const queue = new Queue({ concurrency: this.maxProviders })
queue.addEventListener("failure", (evt) => { … })   // never removed
queue.addEventListener("success", (evt) => { … })   // never removed
queue.addEventListener("idle",    () => { … })      // never removed
…
finally {
  this.removeEventListener("provider", peerAddedToSessionListener)     // removed
[...]

PR #1102, opened 2026-08-09
fix: remove per-request queue listeners in abstract session

Rinse12:
> [!IMPORTANT]
> **Reframed 2026-08-12: this is hygiene, not a leak fix.** The justification I originally gave
> was wrong — see the retraction on #1101. A listener-bearing `EventTarget` *is* collected in
> Gecko; the cycle collector reclaims it. Measured with a `FinalizationRegistry`, a real
> `Queue` with all three of these listeners still attached and nothing else referencing it is
> collected 100% of the time. The queues that *were* being retained were held by a separate
> `main-event` bug, and patching that releases them whether or not this PR is applied.
>
> Retracted specifically: "every queue stays a GC root", the 53.5%-of-live-heap figure, and
> the ~930/h rate.

## What the change is for

`AbstractSession.retrieve()` adds three listeners to a per-request `Queue` and never removes
them, while removing the `provider` and `abort` listeners a few lines away in the same
`finally`. This makes the teardown consistent: what the method registers, the method releases.

It is worth having on its own terms — cleanup should not depend on the surrounding collector
being clever enough to make the omission harmless — but I want to be straight that it does not
fix a measured leak, and you should feel free to close it if you would rather not carry the
extra bookkeeping for that.

## Notes

- The three handlers were inline arrow functions; they are now named consts, which is the only
  reason the diff looks larger than three added lines.
- Event types come from `QueueJobFailure` / `QueueJobSuccess`, already exported from
  `@libp2p/utils`; the queue's job-options type is pulled into a local alias so the listener
  signatures can reuse it.
- No test: with the mechanism claim retracted there is no observable behaviour to pin here
  beyond "the listener list is empty afterwards", which `listenerCount()` already reports.
- `packages/utils` does not typecheck from a clean `npm install` on `main` (unbuilt
  `@helia/interface` siblings); I verified this change introduces no new diagnostics in
  `abstract-session.ts` beyond those already present.

### DIFF
--- a/packages/utils/src/abstract-session.ts
+++ b/packages/utils/src/abstract-session.ts
@@ -6,7 +6,7 @@ import pDefer from 'p-defer'
 import { raceSignal } from 'race-signal'
 import type { BlockBroker, BlockRetrievalOptions, CreateSessionOptions } from '@helia/interface'
 import type { AbortOptions, ComponentLogger, Logger } from '@libp2p/interface'
-import type { Filter } from '@libp2p/utils'
+import type { Filter, QueueJobFailure, QueueJobSuccess } from '@libp2p/utils'
 import type { Multiaddr } from '@multiformats/multiaddr'
 import type { CID } from 'multiformats/cid'
 import type { DeferredPromise } from 'p-defer'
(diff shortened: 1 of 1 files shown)


achingbrain:
Closing as not required

PR #1129, opened 2026-09-14
fix: pass runOnLimitedConnections to the topology and the dial

### LINKED ISSUE #1124: bitswap: runOnLimitedConnections has no effect — two call sites drop the flag
## Summary

`@helia/bitswap` documents `runOnLimitedConnections`, but setting it to `true` does not make bitswap work over a limited (circuit-relay) connection. Two independent places drop the flag, and **each one alone is enough to break it** — patching either has no effect, patching both makes it work.

Versions: `@helia/bitswap` 4.0.11, `helia` 7.1.7, `libp2p` 3.3.8, `@libp2p/circuit-relay-v2` 4.2.11. Two browsers (Chromium), meeting through a circuit relay.

## Where the flag is lost

**1. The registrar topology never notifies about limited connections.**

`src/network.js`, `start()` registers a topology that does not set `notifyOnLimitedConnection`:

```js
const topology = {
  onConnect: (peerId) => { this.safeDispatchEvent('peer:connected', { detail: peerId }) },
  ...
}
```

`libp2p/src/registrar.js` skips exactly that case:

```js
if (connection.limits != null && topology.notifyOnLimitedConnection !== true) { /* skipped */ }
```

So bitswap never learns the peer exists and never asks it for anything.

**2. `sendMessage()` dials without the flag.**

`src/network.js`:

```js
const stream = await this.libp2p.dialProtocol(peerId, BITSWAP_120, options)
```

`options` here is the queue job's options; `this.runOnLimitedConnections` is not merged in. `handle()` and `findProviders()`'s `isDialable()` both honour it — the call that actually asks a peer for a block is the one that does not.

## Measured

Two browser nodes, no direct path possible (no DCUtR service and no `/webrtc` listen address), so the circuit is the only path they can ever have. The connection reports `limits != null`. One node adds a block; the other reads it by CID.

| patch applied | `runOnLimitedConnections` | result |
| --- | --- | --- |
| none | `true` | times out |
| topology only | `true` | times out |
| `sendMessage` only | `true` | times out |
| **both** | `true` | **block arrives** |
| both | `false` (default) | times out |

The last row matters: with both fixed, the default behaviour is unchanged. Nothing starts crossing circuits unless the option asks for it.

Worth noting separately — the bitswap **protocol stream itself opens fine** over the circuit. Dialling `/ipfs/bitswap/1.2.0` by hand with `runOnLimitedConnection: true` succeeds while a read is still timing out, which is what pointed at these two sites rather than at libp2p refusing the protocol.

## Suggested patch

```diff
         const topology = {
+            notifyOnLimitedConnection: this.runOnLimitedConnections,
             onConnect: (peerId) => {
```

```diff
-            const stream = await this.libp2p.dialProtocol(peerId, BITSWAP_120, options);
+            const stream = await this.libp2p.dialProtocol(peerId, BITSWAP_120, {
+                ...options,
+                runOnLimitedConnection: this.runOnLimitedConnections
+            });
```

## Why this matters for browsers

Two browsers behind carrier NAT cannot hole-punch reliably, and the circuit is often the only path they have. Today that means content cannot move between them at all through Helia, while everything else on the same connection — identify, ping, gossipsub, an application protocol declaring `runOnLimitedConnection: true` — works. The option exists to allow that trade; it just does not reach the code that would honour it.

Happy to open a PR with the two-line change and a test if that is useful.

Possibly related, though not the same thing: #253.

NiKrause:
## Title

fix: pass runOnLimitedConnections to the topology and the dial

## Description

`runOnLimitedConnections` reached `handle()` and `findProviders()`'s `isDialable()`, but not the two places that decide whether bitswap runs on a limited connection:

- the topology registered in `start()` did not set `notifyOnLimitedConnection`, so the registrar never reported a peer connected over a limited connection and no wantlist was sent to it
- `sendMessage()` dialled without `runOnLimitedConnection`, so libp2p refused to open the protocol stream

Each one alone is enough to keep bitswap off a limited connection. Both now follow the option. The default is unchanged: nothing starts running on limited connections unless the option asks for it.

Fixes #1124

## Notes & open questions

Tests are four unit tests in `network.spec.ts`, against the existing `stubInterface<Libp2p>`. Each site gets one test with the option on and one with it off. Measured with mocha against the compiled spec, without `--bail`, so every result below is a full count:

| | result |
| --- | --- |
| tests added, fix not applied | 16 passing, 2 failing — exactly the two *"when running on limited connections"* tests |
| fix applied | 18 passing |
| topology site reverted | 17 / 1 — only *"should ask to be notified … when running on limited connections"* |
| dial site reverted | 17 / 1 — only *"should open a stream … when running on limited connections"* |
| topology forced to `true` | 17 / 1 — only *"should not ask to be notified … by default"* |
| dial forced to `true` | 17 / 1 — only *"should not open a stream … by default"* |
| `aegir test -t node` for the package | 70 passing |
| `aegir lint` | clean |

Each test catches exactly its own site, in both directions.

Browser and webworker targets were not run locally. The change is not environment-specific, so I have left those to CI.

End to end: the same two changes have been applied to 4.0.11's build as a `patch-package` patch in [Le-Space/ablage](https://github.com/Le-Space/ablage) since 2026-09-09. Two browsers on a circuit-only connection, as described in the issue: with the option on a block arrives, with it off it does not.

## Change checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


achingbrain (approved):

achingbrain:
Thanks!

PR #1130, opened 2026-09-15
chore: release main

### LINKED ISSUE #1124: bitswap: runOnLimitedConnections has no effect — two call sites drop the flag

achingbrain:
:robot: I have created a release *beep* *boop*
---

<details><summary>bitswap: 4.0.17</summary>

### Bug Fixes

* pass runOnLimitedConnections to the topology and the dial ([#1129](https://github.com/ipfs/helia/issues/1129)) ([bbf05cb](https://github.com/ipfs/helia/commit/bbf05cb7b862cc1a02f5a0fc62412bb141c903cd)), closes [#1124](https://github.com/ipfs/helia/issues/1124)
</details>

<details><summary>interop: 12.1.7</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @helia/bitswap bumped from ^4.0.16 to ^4.0.17
    * helia bumped from ^7.1.12 to ^7.1.13
</details>

<details><summary>helia: 7.1.13</summary>

### Dependencies

* The following workspace dependencies were updated
  * dependencies
    * @helia/bitswap bumped from ^4.0.16 to ^4.0.17
</details>

---
This PR was generated with [Release Please](https://github.com/googleapis/release-please). See [documentation](https://github.com/googleapis/release-please#release-please).


NiKrause:
The one red check here looks like a flake rather than something in this release:

- `test-node (ubuntu-latest, lts/*)` failed in `packages/helia`. `should add a block broker as a factory` ran for 15 minutes, then the process died with `JavaScript heap out of memory` at ~3.9 GB ([job](https://github.com/ipfs/helia/actions/runs/34944452967/job/104301022249)).
- The same test passed in 108 ms in the push run for bbf05cb on `main` ([job](https://github.com/ipfs/helia/actions/runs/34943589549/job/104298327853)), and `test-node` was green there on ubuntu, macOS and Windows.
- This PR is one commit on top of bbf05cb. It changes only versions, the dependency ranges that point at them, the release manifest and changelogs.

Could the job be re-run?

achingbrain:
🤖 Created releases:

- [bitswap-v4.0.17](https://github.com/ipfs/helia/releases/tag/bitswap-v4.0.17)
- [helia-v7.1.13](https://github.com/ipfs/helia/releases/tag/helia-v7.1.13)
- [interop-v12.1.7](https://github.com/ipfs/helia/releases/tag/interop-v12.1.7)

:sunflower:

